PR #11, opened 2024-01-19
Test in-place mutation on mutable inputs

adrhill:


gdalle:
Why not do it for the pushforward too?

adrhill:
Locally, it looks like mutations on `pushforward!` currently fail on the Enzyme backend. Should I fix it in this PR or another one?

On CI, something is still weird since #12.

gdalle:
Feel free to fix it here

PR #19, opened 2024-02-09
Add Documenter key

gdalle:
Is that all I need to do? I also added the key and the secret as specified by DocumenterTools.jl


adrhill:
Looks good to me 👍

Issue #6, opened 2024-01-18
Generality


gdalle:
- Should we support primitives beyond pushforwards / pullbacks ? 
  - Requires making assumptions on argument types
- Should we support higher order derivatives?
  - Requires mixing backends, ergo a non-empty interface
- Should we support argument types beyond arrays / numbers?
  - Makes it harder to reason about mutability and operators
- Should we support multiple arguments?
  - Caused a lot of trouble in AbstractDifferentiation
  - Some backends don't like this anyway (ForwardDiff, ReverseDiff)
- What should be the function signature?
  - `f(x::Array) -> y::Array` as right now (more or less)
  - `f!(y::Array, x::Array) -> z::Nothing` as in [ForwardDiff](https://juliadiff.org/ForwardDiff.jl/stable/user/api/#ForwardDiff.jacobian!)
  - `f!(y::Array, x::Array) -> z::Union{Number,Nothing}` as in [Enzyme](https://enzyme.mit.edu/julia/stable/api/#EnzymeCore.autodiff-Union{Tuple{RABI},%20Tuple{ReturnPrimal},%20Tuple{A},%20Tuple{FA},%20Tuple{ReverseMode{ReturnPrimal,%20RABI},%20FA,%20Type{A},%20Vararg{Any}}}%20where%20{FA%3C:Annotation,%20A%3C:Annotation,%20ReturnPrimal,%20RABI%3C:EnzymeCore.ABI})

adrhill:
## (1) Mutating pushforwards & pullbacks
These are at the core of DifferentiationInterface.
Each backend needs to implement either a `pushforward!`, `pullback!` or both.

(1.1) Mutation of tangent
```julia
pushforward!(dy, backend, f, x, dx) -> (y, dy)
pullback!(dx, backend, f, x, dy) -> (y, dx)
```
* Q: Do we want to support multiple arguments?
* Q: If so, how? Tuples? (See [#53](https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/53))
* Q: What do we understand by mutation? Does this simply provide a "scratch space" in memory for the function or does the result have to be written there?
* Q: Does the `pushforward` abstraction fully cover ForwardDiff and FiniteDiff/erences?

(1.2) Mutation of primal and tangent
```julia
pushforward!(y, dy, backend, f!, x, dx) -> (y, dy)
pullback!(y, dx, backend, f!, x, dy) -> (y, dx)
```
[...]

PR #22, opened 2024-03-01
Add support for Jacobians

adrhill:
This PR adds fallback implementatios for `value_and_jacobian!`, `value_and_jacobian`, `jacobian!`, and `jacobian`.

The implementation dispatches on `AbstractForwardBackend` and `AbstractReverseBackend`:
* `AbstractForwardBackend`: Makes use of `value_and_pushforward!`, which computes JVPs. The pushforward of a standard basis vector $e_i$ computes the $i$-th column of a Jacobian. The Jacobian is therefore assembled column-wise.
* `AbstractReverseBackend`: Makes use of `value_and_pullback!`, which compute VJPs. The pullback of a standard basis vector $e_i$ computes the $i$-th row of a Jacobian. The Jacobian is therefore assembled row-wise.

This requires the implementation of a standard basis vector in `src/unitvector`.
This implementation can be replaced with more performant implementations in the future by specializing on backend types.

To write into pre-allocated Jacobians, type annotations of pullbacks and pushforwards had to be loosened.

Towards #6.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [weakdeps]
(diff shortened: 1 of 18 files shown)


adrhill:
Good news: only Enzyme required specialized `unitvector` and `value_and_pullback!` methods to support inplace mutations of Jacobians!

In both cases, this was due to Enzyme's `Duplicated(x::T, dx::T)` assuming identical types for inputs and outputs of the pullback.

___

One annoyance encountered in this PR is that `value_and_pullback!(dx, ..., x::Real)` assumes that `dx` is scalar and can't be mutated, even though it might in fact be a mutable entry in a pre-allocated Jacobian. A similar case exists for pushforwards on functions with scalar outputs `y`. This required an additional two methods for `_value_and_jacobian!`:
https://github.com/gdalle/DifferentiationInterface.jl/blob/1e40a0d01a76eb3c78141fffbad59bf9bf83add1/src/jacobian.jl#L41-L59

adrhill:
Superseded by https://github.com/gdalle/DifferentiationInterface.jl/pull/23.

PR #24, opened 2024-03-06
Backend-specific utilities (derivative, multiderivative, gradient, jacobian)

gdalle:
**Doc changes**

- Put more info in the README, use it as docs home page
- Introduce the distinction between "primitives" (pushforward, pullback) and "utilities" (derivative, multiderivative, gradient, jacobian)
- Improve documentation, especially for backends
- Avoid exposing the internals of structs (like type parameters), only document constructors and accessors

**Structural changes**

- Add a `custom` type parameter to `AbstractBackend` controling whether custom utilities are used instead of our fallbacks (that call `pushforward` and `pullback`). Implement convenience constructors with the default `custom = true`
- Define a few backend manipulation functions like `handles_input_type`, `handles_output_type`, `autodiff_mode` and `is_custom`
- Add tests for utilities, make sure that they call both custom and fallback versions

**Backend-specific changes**

- Add extensions for PolyesterForwardDiff.jl and Zygote.jl backends
- Implement custom utilities for each backend (lots of code there)
- Activate type stability tests for the backends where they pass
- Get the right pushforward complexity for FiniteDiff.jl backend by derivating the scalar function $t \mapsto f(x + t \delta x)$
- Add finite difference type information to `FiniteDiffBackend` type
- Add `ZygoteBackend()` shortcut with nicer printing

**Benchmark changes**

- Flesh out benchmarks to compare primitives, custom utilities and fallback utilities across all backends
- Benchmark everything on variants of a neural network layer $y = \sigma(wx + b)$


adrhill:
Would be nice if we had benchmarks before blindly merging this.

adrhill:
> Question: should we make these implementations optional with a backend type parameter? 
> Like AbstractBackend{default} which decides whether to use the package fallbacks or the backend implems
> 
> Worry: if we use these specific implems all the time, we never test our fallbacks

Adding a type parameter to all backends sounds good.

Maybe this could also be dealt with through the `extras` argument we planned on adding?

ext/DifferentiationInterfaceZygoteExt.jl
```diff
@@ -1,7 +1,51 @@
...
+"""
+$(TYPEDSIGNATURES)
+"""
+function DI.value_and_jacobian!(
+    jac::AbstractMatrix, backend::ZygoteBackend, f, x::AbstractArray
+)
+    y, new_jac = DI.value_and_jacobian(backend, f, x)
+    jac .= new_jac
+    return y, jac
+end
```
adrhill: Zygote's `withjacobian` would be more efficient as it returns the primal value as well:
https://fluxml.ai/Zygote.jl/dev/utils/#Zygote.withjacobian

adrhill (approved):

ext/DifferentiationInterfaceZygoteExt.jl
gdalle: Yes but it returns a flattened version, as I unfortunately discovered. So I need to run f anyway to get the correct shape

adrhill: Is the primal output of `withjacobian` also flattened?

gdalle: Yes that is what I mean

gdalle:
```julia
julia> Zygote.withjacobian(identity, rand(2, 3)).val
6-element Vector{Float64}:
 0.42579108690107337
 0.9756201877495323
 0.3391154024834867
 0.8593637011392009
 0.6974918525763482
 0.9470770391746074
```

gdalle: https://github.com/FluxML/Zygote.jl/issues/1506

PR #28, opened 2024-03-08
Switch backends to ADTypes.jl

### LINKED ISSUE #15: Depend on ADTypes?
[ADTypes.jl](https://github.com/SciML/ADTypes.jl) contains backend definitions that are very very similar to our own:

https://github.com/SciML/ADTypes.jl/blob/main/src/ADTypes.jl

Maybe we should depend on them. Main challenges are replacing dispatch with traits to:

- distinguish between forward and reverse mode of the same backend (Enzyme, ChainRules once https://github.com/SciML/ADTypes.jl/pull/21 is merged)
- reformulate the custom / fallback dichotomy introduced by #24

gdalle:
**Main idea**

> Replace our backends with the ones from [ADTypes.jl](https://github.com/SciML/ADTypes.jl) to become compatible with the SciML ecosystem.
> ADTypes.jl is used by Lux.jl, Optimization.jl, NonLinearSolve.jl, OrdinaryDiffEq.jl and Turing.jl, among others.

This reduces the number of LOCs in our package, and only moderately complexifies the custom/fallback mechanism.

**Core**

- Add dependency on ADTypes.jl and use their structs to specify backends
- Remove all our backend code, except for the ChainRules one (which ADTypes.jl doesn't have until https://github.com/SciML/ADTypes.jl/pull/21), renamed to `AutoChainRules`
- Implement trait-based dispatch for forward and reverse mode with `autodiff_mode(backend)`, because some backends from ADTypes.jl can do both (mostly `AutoEnzyme`)
- Replace the `custom` type parameter with another trait-based dispatch:
  - `gradient(backend, f, x)` will call the version using custom package functions
  - `gradient(Val(:fallback), backend, f, x)` will call our fallbacks

Note: the `Val`-based traits can be replaced with custom structs if we want

**Tests**

- Make sure that custom and fallback versions are still tested

**Docs**

- Remove the now useless backends page

**Benchmarks**

- Make sure that custom and fallback versions are still benchmarked
- Add pretty printing


benchmark/Project.toml
```diff
@@ -1,9 +1,9 @@
 [deps]
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
-Diffractor = "9f5e2b26-1114-432f-b630-d3fe2085c51c"
```
adrhill: This is due to ADTypes not supporting Diffractor?

src/array_array.jl
```diff
@@ -13,23 +13,39 @@ Compute the primal value `y = f(x)` and the Jacobian matrix `jac = ∂f(x)` of a
...
         DimensionMismatch("Size of Jacobian buffer doesn't match expected size ($ny, $nx)"),
     )
-    return _value_and_jacobian!(jac, backend, f, x, y)
+    return nothing
 end
 
-function _value_and_jacobian!(
+function value_and_jacobian!(
+    ::Val{:fallback},
+    ::Val{:forward},
```
adrhill: Is there a way to somehow combine this with `extras` we plan on adding? Maybe we could have a parametric struct `Extras{Mode,Implem}`.

adrhill: This `Extras` struct could wrap other `extras` arguments. We could have an outer constructor for every backend that sets the mode and implementation.

Something like
```Julia
function value_and_jacobian!(jac, backend, f, x, extras)
    value_and_jacobian!(jac, backend, f, x, Extras(backend, extras))
end
```

This way we could avoid boiler-plate code in all other methods besides the lowest-level `value_and_X!` functions by just forwarding `extras`.

benchmark/Project.toml
gdalle: And Diffractor was failing when I tried to call it through ChainRules. I have opened an issue on the repo, in the meantime I guess we can call it with the AbstractDiff interface as is recommended

src/array_array.jl
gdalle: I don't really like this option because to me the extras should be purely backend specific, and set by the user or by our own "prepare_gradient" methods.
On the other hand, the fallback mechanism is mostly for our internal use and benchmarking, I don't think users should even know about it. After all, if we find that something performs better with our fallbacks, the goal would be to contribute it to the AD package instead of leaving it here.

gdalle: Basically I don't want extras to be somewhere we put configuration or options, cause that would make the API more confusing for the user by mixing different aspects. Most of these configurations go in the backend type anyway (for instance the number of chunks in ForwardDiff or the finite difference type in FiniteDiff). As for the mode it is defined directly and statically by the backend.

gdalle: Finally, note that your approach seems to add one method for each function, so it's mostly the same quantity of "boilerplate" as mine. The only difference is for the utilities where we have to differentiate between forward and reverse mode, but then again you would need to do that too

gdalle: However I do think if we keep my approach we should replace value types with custom structs in the trait mechanism, to avoid accidental piracy

gdalle:
New proposal here:

**Core**

- The trait `Val{:fallback} / Val{:custom}` has become `FallbackImplem() / CustomImplem()` inheriting from `AbstractImplem`
- The trait `Val{:forward} / Val{:reverse}` has become `ForwardMode() / ReverseMode()` inheriting from `AbstractMode`
- Every utility takes `implem::AbstractImplem` as a first argument.
- In the extensions, every utility must only be defined for `implem::CustomImplem`
- Metaprogramming allows every utility to concisely default on `CustomImplem` (this happens in the `custom.jl` file with `@eval`): no boilerplate :sunglasses:
- The argument `implem` is passed down everywhere. This means that if it starts as `CustomImplem()`, it will jump to the custom version every chance it has, and if it starts as `FallbackImplem()`, it will stick to the fallbacks until it reaches a non-utility function (pushforward or pullback).

Basically this is like putting it in `extras` but I prefer keeping both mechanisms separate, because this one does not need to be known by most users 

**Backend-specific**

- Implemented Zygote without ChainRules using just `Zygote.pullback`
- Added a Diffractor extension but it is broken for non-scalar input

adrhill:
I like this a lot. I agree with you on keeping implementations and `extras` separate.

Any thoughts on putting the `implem` and `mode` as last arguments?
This could simplify boilerplate methods like

```Julia
function jacobian!(jac::AbstractMatrix, backend::AbstractBackend, f, x::AbstractArray)
    return last(value_and_jacobian!(jac, backend, f, x))
function jacobian!(
    implem::AbstractImplem,
    jac::AbstractMatrix,
    backend::AbstractADType,
    f,
    x::AbstractArray,
)
    return last(value_and_jacobian!(implem, jac, backend, f, x))
end
```

to a single method
```Julia
function jacobian!(
    jac::AbstractMatrix,
    backend::AbstractADType,
    f,
    x::AbstractArray,
    args...
)
    return last(value_and_jacobian!(jac, backend, f, x, args...))
end
```
and has the additional benefit of making the mutated input the first argument.
This would also forward the `extras` which we plan on adding.

gdalle:
Yeah that's a good idea, I'll do that

adrhill (approved):

PR #36, opened 2024-03-13
Make all backends compatible with all input/output pairs

gdalle:
## Problem

We want all backends to support scalar or array inputs and outputs, but

- Reverse mode Enzyme only supports [scalar or `nothing` output](https://enzymead.github.io/Enzyme.jl/stable/api/#EnzymeCore.autodiff-Union{Tuple{RABI},%20Tuple{ReturnPrimal},%20Tuple{A},%20Tuple{FA},%20Tuple{ReverseMode{ReturnPrimal,%20RABI},%20FA,%20Type{A},%20Vararg{Any}}}%20where%20{FA%3C:Annotation,%20A%3C:Annotation,%20ReturnPrimal,%20RABI%3C:EnzymeCore.ABI})
- ReverseDiff API only supports [array input](https://juliadiff.org/ReverseDiff.jl/api/)

## Solution

- For Enzyme, wrap `f(x)::AbstractArray` into a callable struct `mf!` that does `mf!(y, x) = y .= f(x)`, and differentiate through this struct
- For ReverseDiff, compose `f(x::Number)` with `x -> [x]` and differentiate the result
- Remove `handles_types` machinery

@adrhill is it a good idea?

### DIFF
--- a/ext/DifferentiationInterfaceEnzymeExt/reverse.jl
+++ b/ext/DifferentiationInterfaceEnzymeExt/reverse.jl
@@ -35,7 +35,7 @@ function DI.value_and_pullback!(
 ) where {X<:Number,Y<:AbstractArray}
     y = f(x)
     mf = Mutating(f)
-    _, new_dx = autodiff(Reverse, mf, Const, Duplicated(y, dy), Active(x))
+    _, new_dx = only(autodiff(Reverse, mf, Const, Duplicated(y, copy(dy)), Active(x)))
     return y, new_dx
 end
 
@@ -43,9 +43,11 @@ function DI.value_and_pullback!(
     dx, ::AutoReverseEnzyme, f, x::X, dy::Y, extras::Nothing=nothing
 ) where {X<:AbstractArray,Y<:AbstractArray}
     y = f(x)
-    dx .= zero(eltype(dx))
+    dx_righttype = similar(x)
+    dx_righttype .= zero(eltype(dx_righttype))
     mf = Mutating(f)
-    autodiff(Reverse, mf, Const, Duplicated(y, dy), Duplicated(x, dx))
+    autodiff(Reverse, mf, Const, Duplicated(y, copy(dy)), Duplicated(x, dx_righttype))
+    dx .= dx_righttype
     return y, dx
 end
 
--- a/ext/DifferentiationInterfaceEnzymeExt/reverse.jl
+++ b/ext/DifferentiationInterfaceEnzymeExt/reverse.jl
@@ -1,13 +1,14 @@
 const AutoReverseEnzyme = AutoEnzyme{Val{:reverse}}
 DI.autodiff_mode(::AutoReverseEnzyme) = DI.ReverseMode()
 
-struct Mutating{F}
+# see https://enzymead.github.io/Enzyme.jl/stable/pullbacks/
+
+struct MakeFunctionMutating{F}
     f::F
 end
 
-function (m::Mutating{F})(y::AbstractArray, x) where {F}
-    # see https://enzymead.github.io/Enzyme.jl/stable/pullbacks/
-    y .= m.f(x)
+function (mf!::MakeFunctionMutating)(y::AbstractArray, x)
+    y .= mf!.f(x)
     return nothing
 end
 
@@ -34,8 +35,8 @@ function DI.value_and_pullback!(
     _dx, ::AutoReverseEnzyme, f, x::X, dy::Y, extras::Nothing=nothing
 ) where {X<:Number,Y<:AbstractArray}
     y = f(x)
-    mf = Mutating(f)
-    _, new_dx = only(autodiff(Reverse, mf, Const, Duplicated(y, copy(dy)), Active(x)))
+    mf! = MakeFunctionMutating(f)
+    _, new_dx = only(autodiff(Reverse, mf!, Const, Duplicated(y, copy(dy)), Active(x)))
     return y, new_dx
 end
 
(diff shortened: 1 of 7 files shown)


ext/DifferentiationInterfaceReverseDiffExt.jl
```diff
@@ -34,6 +30,15 @@ function DI.value_and_pullback!(
     return y, dx
 end
 
+function DI.value_and_pullback!(
+    _dx, backend::AutoReverseDiff, f, x::X, dy::Y, extras::Nothing=nothing
+) where {X<:Number,Y}
+    x_array = [x]
+    dx_array = [zero(x)]
```
adrhill: or `dx_array = zero(x_array)`.

ext/DifferentiationInterfaceReverseDiffExt.jl
```diff
@@ -34,6 +30,15 @@ function DI.value_and_pullback!(
...
+    y, dx_array = DI.value_and_pullback!(dx_array, backend, f ∘ only, x_array, dy, extras)
```
adrhill, line 38: Why do we call `only` on the output of `f`?

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
```diff
@@ -1,6 +1,15 @@
 const AutoReverseEnzyme = AutoEnzyme{Val{:reverse}}
 DI.autodiff_mode(::AutoReverseEnzyme) = DI.ReverseMode()
-DI.handles_output_type(::AutoReverseEnzyme, ::Type{<:AbstractArray}) = false
+
+struct Mutating{F}
```
adrhill: Maybe this could be given a more informative name in case it pops up in stack-traces.

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
```diff
@@ -1,6 +1,15 @@
...
+    f::F
+end
+
+function (m::Mutating{F})(y::AbstractArray, x) where {F}
```
adrhill: Why `where {F}`?

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
```diff
@@ -21,6 +30,27 @@ function DI.value_and_pullback!(
...
+    _, new_dx = only(autodiff(Reverse, mf, Const, Duplicated(y, copy(dy)), Active(x)))
+    return y, new_dx
+end
+
+function DI.value_and_pullback!(
+    dx, ::AutoReverseEnzyme, f, x::X, dy::Y, extras::Nothing=nothing
+) where {X<:AbstractArray,Y<:AbstractArray}
+    y = f(x)
+    dx_righttype = similar(x)
+    dx_righttype .= zero(eltype(dx_righttype))
```
adrhill: Why not `dx_righttype = zero(x)` in the first place?

The variable name is also a bit weird, I would assume we always want our computations to use the right types. ;)

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
adrhill: Doesn't Enzyme require `f!(x, y)` instead of `f!(y, x)`?
https://enzymead.github.io/Enzyme.jl/stable/generated/autodiff/#Defining-a-function

Or does it not matter?

ext/DifferentiationInterfaceReverseDiffExt.jl
gdalle: doesn't really matter since `x` is scalar

ext/DifferentiationInterfaceReverseDiffExt.jl
gdalle: no we call `only` on the vector input, to turn it into a scalar that `f` can eat

gdalle: you misread the composition syntax

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
gdalle: `MutatingWrapperForEnzymeReverseBackend`

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
gdalle: to force specialization

gdalle: https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

gdalle: not sure it's useful here tho

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
gdalle: it's the right type to create a `Duplicated` where `x` and `dx` are the same type

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
gdalle: it doesn't matter

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
adrhill: `MakeAllocatingFunctionMutatingForDifferentiationInterfaceEnzymeReverseBackendWrapper` 😜

ext/DifferentiationInterfaceReverseDiffExt.jl
gdalle: but we could use a `FillArray`

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
adrhill: I think the code already expresses that. 🤷‍♂️

ext/DifferentiationInterfaceReverseDiffExt.jl
adrhill: Wouldn't that clash with `Duplicate`?

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
adrhill: Maybe `MakeFunctionMutatingWrapper`?

ext/DifferentiationInterfaceReverseDiffExt.jl
gdalle: I don't think so but let's keep normal arrays for now

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
gdalle: I split the difference with `MakeFunctionMutating`

ext/DifferentiationInterfaceEnzymeExt/reverse.jl
gdalle: renamed to `dx_like_x = zero(x)`

adrhill (approved):

PR #39, opened 2024-03-13
Fix Mermaid diagrams

adrhill:
* Moves edges between subgraphs out of subgraphs. This appears to have stopped Mermaid's left-to-right layouting. Now everything is nice and symmetric again.
* Adds edge annotations for repeated function calls.

## Preview
Screenshot 2024-03-13 at 17 31 26
Screenshot 2024-03-13 at 17 31 37


gdalle (approved):
Ship it!

PR #40, opened 2024-03-14
Put testing utilities in extension

gdalle:
- Define `Scenario` struct and empty test functions in the `DifferentiationInterface.DifferentiationTest` submodule
- Move the rest of the code from `test/utils.jl` and `test/scenarios.jl` to the extension `DifferentiationInterfaceTestExt`, triggered by `ForwardDiff, JET, Random, Test`
- Reduce the depth of nested loops by using `@testset for`
- Add dummy zero backends to test type stability of the fallback structure
- Remove allocation checks cause they would need to be done with BenchmarkTools to be accurate
- Add nice hints for `MethodError`s to load extensions with https://docs.julialang.org/en/v1/base/base/#Base.Experimental.register_error_hint


gdalle:
Missing code cov is due to `@testset`s not being counted, unsure why

PR #45, opened 2024-03-15
Better ambiguity resolution, taped ReverseDiff

gdalle:
**Core**

- Revamp the dispatch with `@eval` in `src/additional_args.jl` to first add `extras=nothing` then `mode=autodiff_mode(backend)`
- That way every method implementation must have `extras` as an arg _without default value_

**Extensions**

- Implement preparation for `gradient` (mutating) and `jacobian` (allocating, mutating) in ReverseDiff
- The `compile` attribute determines whether the tape is compiled first

**Tests**

- Make Aqua ambiguity check happy
- Add compiled ReverseDiff

**Benchmarks**

- Add compiled ReverseDiff


gdalle:
Benchmarks on my computer:

[results.csv](https://github.com/gdalle/DifferentiationInterface.jl/files/14617532/results.csv)

PR #71, opened 2024-03-19
Benchmarks

gdalle:
**Core**

- Add benchmarking and allocation-testing functions inside `DifferentiationTest`
- The common entry point is now

```julia
"""
    test_operators(
        backends, [operators, scenarios];
        correctness, type_stability, benchmark, allocations,
        input_type, output_type,
        first_order, second_order, allocating, mutating,
        excluded,
    )

Cross-test a list of `backends` for a list of `operators` on a list of `scenarios.`

Return `nothing`, except when `benchmark=true`.

# Default arguments

- `operators`: defaults to all of them
- `scenarios`: defaults to a set of default scenarios

# Keyword arguments

- `correctness=true`: whether to compare the differentiation results with those given by ForwardDiff.jl
- `type_stability=true`: whether to check type stability with JET.jl
- `benchmark=false`: whether to run and return a benchmark suite with Chairmarks.jl
- `allocations=false`: whether to check that the benchmarks are allocation-free
- `input_type=Any`: restrict scenario inputs to subtypes of this
- `output_type=Any`: restrict scenario outputs to subtypes of this
- `first_order=true`: consider first order operators
- `second_order=true`: consider second order operators
- `allocating=true`: consider operators for allocating functions
- `mutating=true`: consider operators for mutating functions
- `excluded=Symbol[]`: list of excluded operators
"""
```

**Extensions**

- Revamp the Chairmarks extension to return a homemade benchmark suite instead of just testing allocations
- Add a Chairmarks + DataFrames extension to process this suite into a table

**Tests**

- Better nest test sets
- Add dependencies to make each test file self sufficient


gdalle:
Overview of the resulting dataframe:

image

image

PR #73, opened 2024-03-19
Remove benchmark suite

gdalle:


adrhill:
Why do you want to remove it completely?

gdalle:
Because it's outdated anyway, and there is a new benchmarking infrastructure in place. Since the package is very much in flux at the moment, we don't expect to have optimal performance yet, so running this in CI is rather useless anyway

gdalle:
Do you wanna keep it?

adrhill:
Once our benchmarking infrastructure settles, it would be nice to be able to run CI benchmarks on it. 
Is ChairMarks compatible with BenchmarkCI?

gdalle:
No, that's the issue. Chairmarks does not yet have a notion of benchmark suite / group (which is why I recoded it myself), let alone predefined CI scripts

gdalle:
Luckily, the way it is coded, the benchmarking part is very isolated and easy enough to switch from Chairmarks to BT if we want

Issue #35, opened 2024-03-13
Structured arrays


gdalle:
MWE by @mcabbott

```julia
julia> backend = ADTypes.AutoForwardDiff();  # from docs example

julia> DifferentiationInterface.value_and_gradient(backend, f, [1., 2., 3.])
(14.0, [2.0, 4.0, 6.0])

julia> DifferentiationInterface.value_and_gradient(backend, f, LinRange(1,3,3))
(14.0, [2.0, 4.0, 6.0])

julia> DifferentiationInterface.value_and_gradient(backend, f, Symmetric(ones(3,3)))
ERROR: ArgumentError: Cannot set a non-diagonal index in a symmetric matrix
```

The issue is that `value_and_gradient` creates a `grad` object with `similar(x)` and then tries to mutate it by calling `value_and_gradient!`. This fails for a `Symmetric` matrix.

gdalle:
The package is gonna be agnostic about tangent / cotangent structure, but we can try to avoid some of these bugs with a more careful fallback structure (like avoiding mutation when we can).

gdalle:
See the recent discussion on Enzyme:
- https://github.com/EnzymeAD/Enzyme.jl/issues/1334

gdalle:
The question of sparse arrays is also relevant

adrhill:
> we can try to avoid some of these bugs with a more careful fallback structure (like avoiding mutation when we can). 

What would these fallbacks dispatch on? It sounds like this would require adding traits for each array type in the Julia ecosystem. 
Or should we change our existing `AbstractArray` type annotations to `Array` and be more conservative / non-mutating with `AbstractArray`?

gdalle:
Yeah this is a limitation of ForwardDiff actually. We don't have any grasp on how the backends handle structure, and we cannot work without at least mutation

gdalle:
Closing it cause there isn't much we can do

mcabbott:
It's possible that ForwardDiff should call 3-arg `similar(x, T, axes(x))` to strip structure. Without thinking too hard about the mathematics, this appears to match Zygote (where we did think hard):
```julia
julia> using ForwardDiff, Zygote

julia> f(x) = sum(abs2, x);

julia> x = Symmetric(rand(Int8, 3, 3) .+ 0.0)
3×3 Symmetric{Float64, Matrix{Float64}}:
 -35.0   -52.0  -62.0
 -52.0  -113.0  -75.0
[...]

PR #74, opened 2024-03-19
Default to three-argument `similar` for pre-allocations

adrhill:
As discussed in #35, this can be used to strip array structure.
`mysimilar` is only used in the fallbacks for `value_and_pushforward` and `value_and_pullback` to allocate pushforward and pullback arrays in memory, so this doesn't solve existing issues with ForwardDiff's `gradient!` implementation.


gdalle:
Can't hurt!

PR #76, opened 2024-03-20
Better backend docs

gdalle:
- Add functions to check support for mutation and second-order differentiation
- Use them to generate tables of backend functionalities, including backend combinations for hessians


gdalle:
Output preview (the table is scrollable):

image

PR #77, opened 2024-03-20
HVP for everyone

gdalle:
**Core**

- Define `gradient_and_hessian_vector_product` for every second order backend (except symbolic ones), mutation-compatible or not
- Better `supports_mutation` and `support_hessian` utilities, checking correct result (that way `FiniteDiff` over `FiniteDiff` fails)

**Tests**

- Add second-order tests for more backends, especially `SecondOrder`
- Add correctness, stability and allocation tests for `gradient_and_hessian_vector_product`
- Better scenario display in test set

**Extensions**

- Enzyme: Make sure to convert `dx` everywhere before building `Duplicated(x, dx)`


gdalle:
Enzyme is crapping all over the test logs: https://github.com/EnzymeAD/Enzyme.jl/issues/1354

PR #113, opened 2024-03-27
Pushforwards and pullbacks for everyone

gdalle:
**Core**

- Implement pushforward based on pullback and pullback based on pushforward => if you forget both, you get a `StackOverflowError`
- Make `derivative` a simple pushforward of 1 and `gradient` a simple pullback of 1
- Replace traits `...Supported` with `...Fast`, because some backends may support pushforwards and pullbacks
- Replace functions `supports_...` with `..._performance`

**Tests**

- Split testing from benchmarking
- Remove allocation tests
- Test all scenarios on pushforwards and pullbacks
- Add optional logging to test function instead of test file

### DIFF
--- a/ext/DifferentiationInterfaceFiniteDifferencesExt/DifferentiationInterfaceFiniteDifferencesExt.jl
+++ b/ext/DifferentiationInterfaceFiniteDifferencesExt/DifferentiationInterfaceFiniteDifferencesExt.jl
@@ -3,7 +3,7 @@ module DifferentiationInterfaceFiniteDifferencesExt
 using ADTypes: AutoFiniteDifferences
 import DifferentiationInterface as DI
 using FillArrays: OneElement
-using FiniteDifferences: FiniteDifferences, jvp
+using FiniteDifferences: FiniteDifferences, jvp, j′vp
 using LinearAlgebra: dot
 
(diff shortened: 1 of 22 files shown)


ext/DifferentiationInterfaceFiniteDifferencesExt/DifferentiationInterfaceFiniteDifferencesExt.jl
```diff
@@ -19,4 +19,15 @@ function DI.value_and_pushforward(
     return y, jvp(backend.fdm, f, (x, dx))
 end
 
+#=
+# TODO: why does this fail?
```
gdalle, line 23: @adrhill this is the function we need for NNLib, so I'd be interested in the answer

adrhill: How/why does it fail?

gdalle: Fixed in #114, it was due to https://github.com/JuliaDiff/FiniteDifferences.jl/issues/234 among other things

PR #120, opened 2024-03-28
Separate testing scenarios for each operator

adrhill:
Opening this as a draft PR for feedback before updating the default scenarios in `src/scenarios`.


gdalle:
**Testing**

- Create scenario types specific to each operator
- Document most of it in the docstring of `AbstractScenario`
- Add minimal type constraints in the scenario type parameters
- Write the correct constructors
- Apply them to the default, GPU, static and component test suites
- Update the correctness, type stability, call count, sparsity and benchmark subroutines
- Pass the `ref_backend` from above if needed to override scenario-specific reference
- Replace `@info` with ProgressMeter.jl to avoid flushing the terminal

**Docs**

- Create TOC sections on the sidebar
- Separate core and testing API references
- Separate extensions from backends

PR #141, opened 2024-04-05
Tapir mutation, take 2

gdalle:
**Extensions**

- Make Tapir compatible with mutating functions


gdalle:
@willtebbutt I tried following your advice from https://github.com/withbayes/Tapir.jl/issues/113#issuecomment-2036718992 but now I get a new error:

```julia
julia> test_differentiation(
           all_backends; second_order=false, logging=get(ENV, "CI", "false") == "false"
       );
Correctness: Error During Test at /home/gdalle/Work/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterfaceTest/src/test_differentiation.jl:94
  Got exception outside of a @test
  MethodError: Cannot `convert` an object of type NoTangent to an object of type Tapir.Tangent{@NamedTuple{a::Vector{NoTangent}}}
  function:  num_to_arr!
  Closest candidates are:
    convert(::Type{T}, ::T) where T
     @ Base Base.jl:84
    (::Type{Tapir.Tangent{Tfields}} where Tfields<:NamedTuple)(::Any)
     @ Tapir ~/.julia/packages/Tapir/BqxEi/src/tangents.jl:53
  
  Stacktrace:
    [1] setindex!(A::Vector{Tapir.Tangent{@NamedTuple{a::Vector{NoTangent}}}}, x::NoTangent, i1::Int64)
      @ Base ./array.jl:1021
    [2] setindex!(b::Base.RefArray{Tapir.Tangent{@NamedTuple{…}}, Vector{Tapir.Tangent{…}}, Nothing}, x::NoTangent)
      @ Base ./refpointer.jl:185
    [3] map
      @ ./tuple.jl:322 [inlined]
    [4] Pullback
      @ ~/.julia/packages/Tapir/BqxEi/src/interpreter/s2s_reverse_mode_ad.jl:632 [inlined]
    [5] value_and_pullback!!(f!::Function, y::Vector{…}, dx::Float64, ::AutoTapir, x::Float64, dy::FillArrays.OneElement{…}, extras::DifferentiationInterfaceTapirExt.TapirMutatingPullbackExtras{…})
      @ DifferentiationInterfaceTapirExt ~/Work/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/mutating.jl:44
```

It is very confusing to me because there are no named tuples anywhere in my testing code

willtebbutt:
Aha, right -- it looks like `f!` is a closure in some cases, which means that there are some parameters in it. Specifically, it looks like it's a matrix or vector of `Int`s, depending on the function. Ideally the `tangent_type(typeof(f!))` of such a function would be `NoTangent`, but it's currently `Tapir.Tangent{@NamedTuple{a::Vector{NoTangent}}}`. This is where the `NamedTuple` is coming from -- Tapir sees `f!` as a struct with a field containing a vector of non-differentiable things.

To deal with this, we should just relax the constraint that the tangent type of `f!` must be `NoTangent` (which this method wasn't actually testing), and generate a `zero_tangent` for `f!`. The following works locally for me:
```julia
function DI.value_and_pullback!!(
    f!::Tf, y, dx, ::AutoTapir, x, dy, extras::TapirMutatingPullbackExtras
) where {Tf}

    dy_righttype = convert(tangent_type(typeof(y)), dy)
    dx_righttype = convert(tangent_type(typeof(x)), dx)

    # We want the VJP, not VJP + dx, so I'm going to zero-out `dx`. `set_to_zero!!` has the advantage
    # that it will also replace any immutable components of `dx` to zero.
    dx_righttype = set_to_zero!!(dx_righttype)

    # We want `dy` to correspond to the cotangent of `y` _after_
    # running the forwards-pass, so I'm going to take a copy, and zero-out the original.
    dy_righttype_backup = copy(dy_righttype)
    dy_righttype = set_to_zero!!(dy_righttype)

    # Mutate a copy of `y`, so that we can run the reverse-pass later on.
    y_copy = copy(y)

    df! = zero_tangent(f!)

    # Run the forwards-pass.
    out, pb!! = extras.rrule(
        CoDual(f!, df!), CoDual(y_copy, dy_righttype), CoDual(x, dx_righttype)
    )

    # Verify that the output is non-differentiable.
    @assert tangent(out) == NoTangent()

    # Set the cotangent of `y` to be equal to the requested value.
    dy_righttype = increment!!(dy_righttype, dy_righttype_backup)

    # Record the state of `y` before running the reverse-pass.
    y = copy!(y, y_copy)

    # Run the reverse-pass.
    _, _, new_dx = pb!!(NoTangent(), df!, dy_righttype, dx_righttype)

    return y, new_dx
end
```

There are some remaining failures, but they look like they're to do with split mode?

gdalle:
This works!!! Thank you. Indeed, in the general case f! might be a closure or callable struct, but we never want to differentiate with respect to its contents (at least for now).

The spurious copies are not a problem, I can preallocate the necessary memory as part of the preparation step, along with computing the rule

PR #165, opened 2024-04-09
Choose sparsity detector

gdalle:
**Extensions**

- Modify SparseDiffTools extension to check if the backend has a `sparsity_detector` field and pick it instead of `SymbolicsSparsityDetection()`


Vaibhavdixit02 (commented):
When called with `prepare_hessian` SparsityConnectivityTracer choice throws

```
ERROR: MethodError: no method matching dual_definition_retval(::Val{…}, ::SparseConnectivityTracer.Tracer, ::Float64, ::ForwardDiff.Partials{…})
```

gdalle:
I think that's a missing method on the SparseConnectivityTracer side, opened an issue to keep track.
Does it work with other backends? I think `AutoSparseForwardDiff` over `AutoFiniteDiff` should work for instance

PR #167, opened 2024-04-10
Remove reference to BangBang

gdalle:
Amend the warning in the docs to avoid suggesting that the convention is exactly the same as BangBang.jl's

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -43,7 +43,7 @@ DifferentiationInterfaceZygoteExt = "Zygote"
 [compat]
 ADTypes = "0.2.7"
 ChainRulesCore = "1.23.0"
-Diffractor = "0.2.6 - 0.2.7"
+Diffractor = "=0.2.6"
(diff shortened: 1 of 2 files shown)


gdalle:
Note however that BangBang.jl itself does not guarantee that the input will be mutated whenever possible:
```julia
julia> using BangBang, StaticArrays

julia> src = [2.0, 3.0];

julia> dst_before = zero(src);

julia> dst_after = broadcast!!(abs2, dst_before, src)
2-element Vector{Float64}:
 4.0
 9.0

julia> dst_before  # it was mutated
2-element Vector{Float64}:
 4.0
 9.0

julia> src_static = MVector(2.0, 3.0);

julia> dst_before_static = zero(src_static);

julia> dst_after_static = broadcast!!(abs2, dst_before_static, src_static)
2-element MVector{2, Float64} with indices SOneTo(2):
 4.0
 9.0

julia> dst_before_static # it wasn't mutated but could have been
2-element MVector{2, Float64} with indices SOneTo(2):
 0.0
 0.0
```

PR #186, opened 2024-04-17
Improve Enzyme in forward mode with chunk size

### LINKED ISSUE #119: Enzyme utilities are type unstable in forward mode
`gradient` and `jacobian` make `JET.@test_opt` angry

We probably need to provide the number of chunks with preparation

gdalle:
**Core**

- Add a `pick_chunksize` heuristic lifted from ForwardDiff.jl

**Extensions**

- Enzyme forward: Use this chunk size selector to prepare `gradient` and `jacobian`
- Enzyme reverse: Try to do the same for `jacobian` but I ran into https://github.com/EnzymeAD/Enzyme.jl/issues/1391

**Tests**

- Add `AutoEnzyme(Enzyme.Forward)` to the type-stable backends


adrhill:
In the long run, would it make sense to add a `backend` argument for dispatch to `pick_chunksize`? Or is the heuristic generally valid?

gdalle:
It can't hurt. But at the moment only ForwardDiff and Enzyme handle chunks, and based on the slack discussion the optimal chunk length is about the same for both (circa SIMD length)

PR #207, opened 2024-04-24
Generate table of overloads in docs

### LINKED ISSUE #115: The big table of overloads
This table recaps the features of each extension, with respect to high-level operators.
Each cell can have four values

- ❌: the backend does not support this operator
- ❔: the backend supports this operator but our extension doesn't call it yet / doesn't handle preparation
- ✅: our extension calls the backend operator and handles preparation if possible

| backend              | derivative | gradient | jacobian | hvp | hessian |
| :------------------- | ---------- | -------- | -------- | --- | ------- |
| ChainRulesCore       | ❌          | ❌        | ❌        | ❌   | ❌       |
| Diffractor           | ❌          | ❌        | ❌        | ❌   | ❌       |
| Enzyme (forward)     | ❌          | ✅        | ✅        | ❌   | ❌       |
| Enzyme (reverse)     | ❌          | ✅        | [❔](https://github.com/EnzymeAD/Enzyme.jl/issues/1391)        | ❌   | ❌       |
| FastDifferentiation  | ✅          | ✅        | ✅        | ✅   | ✅       |
| FiniteDiff           | ✅          | ✅        | ✅        | ❌   | ✅       |
| FiniteDifferences    | ❌          | ✅        | ✅        | ❌   | ❌       |
| ForwardDiff          | ✅          | ✅        | ✅        | ❌   | ✅       |
| PolyesterForwardDiff | ❌          | ✅        | ✅        | ❌   | ❌       |
| ReverseDiff          | ✅          | ✅        | ✅        | ❌   | ✅       |
| SparseDiffTools      | ❌          | ❌        | ✅        | ❌   | [❔](https://github.com/gdalle/DifferentiationInterface.jl/issues/63#issuecomment-2034324854)       |
| Symbolics            | ✅          | ✅        | ✅        | ❌   | ✅       |
| Tapir                | ❌          | [❔](https://github.com/withbayes/Tapir.jl/issues/116)        | ❌        | ❌   | ❌       |
| Tracker              | ❌          | ✅        | ❌        | ❌   | ❌       |
| Zygote               | ❌          | ✅        | ❌        | ❌   | ✅       |

adrhill:
Closes #115.

Generates more granular tables than issue #115.
Inside of the tables, checkmarks link to the respective source code.

Caveat: this doesn't include manual links to open issues.

### Preview
~~[Table of overloads.pdf](https://github.com/gdalle/DifferentiationInterface.jl/files/15097057/Table.of.overloads.pdf)~~

### DIFF
--- a/DifferentiationInterface/docs/src/overloads.md
+++ b/DifferentiationInterface/docs/src/overloads.md
@@ -1,11 +1,11 @@
 # Table of overloads
 
 This table recaps the features of each extension, with respect to high-level operators.
-Each cell can have three values
+Each cell can have three values:
 
 - ❌: the backend does not support this operator
 - ✅: our extension calls the backend operator and handles preparation if possible
-- NA: The operator is not available (e.g. because mutation on scalar function outputs isn't possible).
+- NA: the operator is not available (e.g. because mutation on scalar function outputs isn't possible).
 
 Checkmarks (✅) are clickable and link to the source code.
 
@@ -13,8 +13,18 @@ Checkmarks (✅) are clickable and link to the source code.
 using ADTypes
 using DifferentiationInterface
 using DifferentiationInterface: backend_string
-import Markdown
-import Diffractor, Enzyme, FastDifferentiation, FiniteDiff, FiniteDifferences, ForwardDiff, PolyesterForwardDiff, ReverseDiff, Tapir, Tracker, Zygote
+using Markdown: Markdown
+using Diffractor: Diffractor
+using Enzyme: Enzyme
+using FastDifferentiation: FastDifferentiation
+using FiniteDiff: FiniteDiff
+using FiniteDifferences: FiniteDifferences
+using ForwardDiff: ForwardDiff
+using PolyesterForwardDiff: PolyesterForwardDiff
+using ReverseDiff: ReverseDiff
+using Tapir: Tapir
+using Tracker: Tracker
+using Zygote: Zygote
 
 ext_module(ext::Symbol) = Base.get_extension(DifferentiationInterface, ext)
 
@@ -23,11 +33,11 @@ function all_backends_and_extensions()
         (AutoDiffractor(), ext_module(:DifferentiationInterfaceDiffractorExt)),
         (AutoEnzyme(; mode=Enzyme.Forward), ext_module(:DifferentiationInterfaceEnzymeExt)),
         (AutoEnzyme(; mode=Enzyme.Reverse), ext_module(:DifferentiationInterfaceEnzymeExt)),
-        (AutoFastDifferentiation(), ext_module(:DifferentiationInterfaceFastDifferentiationExt)),
+        (AutoFastDifferentiation(), ext_module(:DifferentiationInterfaceFastDifferentiationExt),),
         (AutoFiniteDiff(), ext_module(:DifferentiationInterfaceFiniteDiffExt)),
-        (AutoFiniteDifferences(; fdm=FiniteDifferences.central_fdm(3, 1)), ext_module(:DifferentiationInterfaceFiniteDifferencesExt)),
+        (AutoFiniteDifferences(; fdm=FiniteDifferences.central_fdm(3, 1)),ext_module(:DifferentiationInterfaceFiniteDifferencesExt),),
         (AutoForwardDiff(), ext_module(:DifferentiationInterfaceForwardDiffExt)),
-        (AutoPolyesterForwardDiff(; chunksize=1), ext_module(:DifferentiationInterfacePolyesterForwardDiffExt)),
+        (AutoPolyesterForwardDiff(; chunksize=1),ext_module(:DifferentiationInterfacePolyesterForwardDiffExt),),
         (AutoReverseDiff(), ext_module(:DifferentiationInterfaceReverseDiffExt)),
         (AutoTapir(), ext_module(:DifferentiationInterfaceTapirExt)),
         (AutoTracker(), ext_module(:DifferentiationInterfaceTrackerExt)),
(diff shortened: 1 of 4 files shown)


gdalle:
bro that rocks

gdalle:
Why is the table not in alphabetical order?
Also, FastDifferentiation is the only backend for which _everything_ is overloaded, so it should be all green?

DifferentiationInterface/docs/src/overloads.md
```diff
@@ -0,0 +1,93 @@
...
+
+    n = length(ms)
+    n == 0 && return "❌"
+    n == 1 && return "[✅]($(Base.url(only(ms))))"
+    return "✅"
+end
+
+io = IOBuffer()
+
+for (backend, ext) in all_backends_and_extensions()
```
gdalle: sort alphabetically?

DifferentiationInterface/docs/src/overloads.md
```diff
@@ -0,0 +1,93 @@
...
+
+    # Subsection title
+    println(io, "## $bname")
+    # Table header
+    println(io, "| Operator | `operator(f, backend, x, ...)`| `operator(f!, y, backend, x, ...)` |")
+    println(io, "|:---|:---:|:---:|")
+    # Table contents
+    for op in operators
+        println(io, "| `$op` | ", 
+            method_overloaded(op, (Any, btype, Any, Any), ext),
```
gdalle: I think this might be missing some overloads because the typing is more specific than `Any`

gdalle (changes requested):

adrhill:
> Why is the table not in alphabetical order?

If you remove the `value_and_` functions, it is.
I can swap `operator!` and `operator` if you prefer.

> Also, FastDifferentiation is the only backend for which everything is overloaded, so it should be all green?

Good catch, that's a bug.

DifferentiationInterface/docs/src/overloads.md
adrhill: Aren't the backends already in alphabetical order?

DifferentiationInterface/docs/src/overloads.md
adrhill: I think it's due to `op!` having a different argument count than `op`.

DifferentiationInterface/docs/src/overloads.md
gdalle: Right, my bad

DifferentiationInterface/docs/src/overloads.md
gdalle: No, it's also because the method overloads won't accept Any in most places, for instance the extras have specific types, sometimes x and y are restricted to numbers or arrays, etc

gdalle: Basically you wanna check if for operator op, there is a method whose signature contains a backend argument more specific than the union of the backend type and its sparse counterpart

gdalle:
I would put more stuff in a single line, like all four variants of gradient for example

adrhill:
Does this look correct for FastDifferentiation?

image

adrhill:
> I would put more stuff in a single line, like all four variants of gradient for example

What would you call the columns?

DifferentiationInterface/docs/src/overloads.md
adrhill: The types are forwarded to `methods(f, (types), extension_module)`, which I think does the opposite:

```Julia
julia> methods(atan, (Float32, Float32))
# 1 method for generic function "atan" from Base:
 [1] atan(y::T, x::T) where T<:Union{Float32, Float64}
     @ Base.Math special/trig.jl:566

julia> methods(atan, (Any, Any))
# 8 methods for generic function "atan" from Base:
 [1] atan(a::Float16, b::Float16)
     @ Base.Math math.jl:1568
 [2] atan(y::BigFloat, x::BigFloat)
     @ Base.MPFR mpfr.jl:812
 [3] atan(::Missing, ::Missing)
     @ Base.Math math.jl:1584
 [4] atan(::Missing, ::Number)
     @ Base.Math math.jl:1586
 [5] atan(y::T, x::T) where T<:Union{Float32, Float64} # <--- (Float32, Float32)
     @ Base.Math special/trig.jl:566
 [6] atan(y::T, x::T) where T<:AbstractFloat
     @ Base.Math math.jl:875
 [7] atan(y::Real, x::Real)
     @ Base.Math math.jl:874
 [8] atan(::Number, ::Missing)
     @ Base.Math math.jl:1585
```

gdalle:
No, it also has HVP. The source of truth is the GitHub issue

DifferentiationInterface/docs/src/overloads.md
gdalle: Maybe try a top down approach, take all methods and filter by backend only

gdalle:
> What would you call the columns?

For each backend, one table for 1-arg functions `f` and one table for 2-arg functions `f!`

In each row, a single operator like `:gradient`, and four columns with headers:

- `op`
- `op!`
- `val_and_op`
- `val_and_op!`

Alternately, the two tables could be brought into one

adrhill:
The problem with `hvp!` and `hvp` was the missing vector argument. 
There are too many edge cases, so now I manually specify operator function signatures for both `f(x)` and `f!(y, x)`.

### Updated preview
~~[Table of overloads.pdf](https://github.com/gdalle/DifferentiationInterface.jl/files/15098038/Table.of.overloads.pdf)~~

gdalle:
> The problem with hvp! and hvp was the missing vector argument.

Beware, `pushforward` and `pullback` also take one

adrhill:
Round 3: ~~[Table of overloads.pdf](https://github.com/gdalle/DifferentiationInterface.jl/files/15098808/Table.of.overloads.pdf)~~

DifferentiationInterface/docs/src/overloads.md
adrhill: This is what I'm already doing:
```Julia
ms = methods(f, argtypes, ext) # only look at methods in matching package extension

n = length(ms)
n == 0 && return "❌"
n == 1 && return "[✅]($(Base.url(only(ms))))"
return "[✅]($(Base.url(first(ms))))" # Optional TODO: return all URLs?
```

gdalle:
It looks better at first glance, I'll check in detail tomorrow! We probably don't need the 2-arg table for backends that don't support these functions, if that's not too much trouble to remove

adrhill:
Good call, I've now filtered them. It makes the tables a lot more readable.
I've also added a couple of sentences to explain to people what they are looking at.

Latest build: [Table of overloads.pdf](https://github.com/gdalle/DifferentiationInterface.jl/files/15101560/Table.of.overloads.pdf)

gdalle (approved):

PR #215, opened 2024-04-25
Update backend documentation

adrhill:
* summarizes all `check_*` functions in one table
* adds examples on how to use ADTypes, e.g. to define forward- and reverse-mode in Enzyme

Preview: [Backends.pdf](https://github.com/gdalle/DifferentiationInterface.jl/files/15110923/Backends.pdf)


gdalle:
This is amazing! Maybe replace "mutation" with "two arguments" to keep terminology coherent throughout?

gdalle (approved):

PR #216, opened 2024-04-25
Simplify tutorial and README

adrhill:
* replace `@repl` blocks by `@example` blocks to reduce visual noise (fewer `julia>`, `;` and newlines)
* this is me being opinionated, but I opted for `using ADTypes` instead of `import ADTypes`
* use several backends in README example to make the point of this package overly obvious
* minor additions to tutorial

### DIFF
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -28,7 +28,7 @@ This package provides a backend-agnostic syntax to differentiate functions of th
 
 ## Compatibility
 
-We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl) `v1.0`:
+We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl) v1.0:
 
 - [ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl)
 - [Diffractor.jl](https://github.com/JuliaDiff/Diffractor.jl)
@@ -74,7 +74,7 @@ import ForwardDiff, Enzyme, Zygote          # import automatic differentiation b
 
 f(x) = sum(abs2, x)
 
-x = [1., 2., 3.]
+x = [1.0, 2.0, 3.0]
 
 value_and_gradient(f, AutoForwardDiff(), x) # returns (14.0, [2.0, 4.0, 6.0]) using ForwardDiff.jl
 value_and_gradient(f, AutoEnzyme(),      x) # returns (14.0, [2.0, 4.0, 6.0]) using Enzyme.jl
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -79,7 +79,7 @@ value_and_gradient(f, AutoEnzyme(),      x) # returns (14.0, [2.0, 4.0, 6.0]) us
 value_and_gradient(f, AutoZygote(),      x) # returns (14.0, [2.0, 4.0, 6.0]) using Zygote.jl
 ```
 
-For more performance, take a look at the [DifferentiationInterface tutorial](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/dev/tutorial/).
+For more performance, take a look at the [DifferentiationInterface tutorial](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/tutorial/).
 
 ## Related packages
 
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -20,7 +20,7 @@ This package provides a backend-agnostic syntax to differentiate functions of th
 
 ## Features
 
-- First- and second-order operators
+- First- and second-order operators (gradients, Jacobians, Hessians and [more](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/overview/))
 - In-place and out-of-place differentiation
 - Preparation mechanism (e.g. to create a config or tape)
 - Thorough validation on standard inputs and outputs (numbers, vectors, matrices)
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -50,9 +50,7 @@ We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/
 To install the stable version of the package, run the following code in a Julia REPL:
 
 ```julia
-julia> using Pkg
-
-julia> Pkg.add("DifferentiationInterface")
+julia> ]add DifferentiationInterface
 ```
 
 To install the development version, run this instead:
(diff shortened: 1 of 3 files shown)


DifferentiationInterface/README.md
```diff
@@ -50,9 +50,7 @@ We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/
 To install the stable version of the package, run the following code in a Julia REPL:
 
 ```julia
-julia> using Pkg
-
-julia> Pkg.add("DifferentiationInterface")
+julia> ]add DifferentiationInterface
```
gdalle: I don't like this because you can't copy-paste it into the REPL, you have to type it yourself, unlike the other one (when you copy-paste a block of REPL examples, `julia>` prompts are ignored)

DifferentiationInterface/docs/src/tutorial.md
```diff
@@ -6,45 +6,45 @@ CurrentModule = Main
...
 We present a typical workflow with DifferentiationInterface.jl and showcase its potential performance benefits.
 
-```@repl tuto
-using DifferentiationInterface
-import ADTypes, ForwardDiff, Enzyme
-using BenchmarkTools
+```@example tuto
+using DifferentiationInterface, ADTypes
+
+import ForwardDiff, Enzyme  # ⚠️ import the backends you want to use ⚠️
```
gdalle, line 12: maybe we could explain why we avoid using here (because there will be name conflicts with DI)

gdalle: also we could do an example of what happens when the backend is not loaded, now that there is a cool error message

DifferentiationInterface/docs/src/tutorial.md
```diff
@@ -58,13 +58,14 @@ Not bad, but you can do better.
 
 ## Overwriting a gradient
 
-Since you know how much space your gradient will occupy, you can pre-allocate that memory and offer it to AD.
+Since you know how much space your gradient will occupy (the same as your input `x`), you can pre-allocate that memory and offer it to AD.
 Some backends get a speed boost from this trick.
 
-```@repl tuto
+```@example tuto
 grad = zero(x)
```
gdalle: should we use `similar`?
- pros: more efficient

DifferentiationInterface/docs/src/tutorial.md
```diff
@@ -76,24 +77,26 @@ More precisely, our convention is that _every positional argument between the fu
 
 For some reason the in-place version is not much better than your first attempt.
 However, it has one less allocation, which corresponds to the gradient vector you provided.
-Don't worry, you're not done yet.
+Don't worry, we can get even more performance.
```
gdalle: I'm careful to distinguish between "we" (Adrian Hill and I) and "you" (the user) in the whole docs, what do you think?

DifferentiationInterface/docs/src/tutorial.md
```diff
@@ -6,45 +6,45 @@ CurrentModule = Main
...
+f(x) = sum(abs2, x)
 
-```@repl tuto
-x = [1.0, 2.0, 3.0];
+x = [1.0, 2.0, 3.0]
+nothing # hide
 ```
 
 To compute its gradient, we need to choose a "backend", i.e. an AD package that DifferentiationInterface.jl will call under the hood.
 Most backend types are defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl) and re-exported by DifferentiationInterface.jl.
```
gdalle, line 33: We should remove that since re-exports are no longer a thing

gdalle (changes requested):

DifferentiationInterface/docs/src/tutorial.md
adrhill: > maybe we could explain why we avoid using here (because there will be name conflicts with DI)

Good point!

> also we could do an example of what happens when the backend is not loaded, now that there is a cool error message

I like the idea, but I would let users discover this on their own (if they need to see it). 😉

DifferentiationInterface/docs/src/tutorial.md
adrhill: Sure, my mistake!

DifferentiationInterface/docs/src/tutorial.md
adrhill: Let's wait for #217 before merging this.

adrhill:
I will make some more changes to this once #217 is merged.

gdalle:
it is merged, you can track all the places in the docs where ADTypes is used and remove them

adrhill:
Ready for review!

adrhill:
@gdalle A test on coloring currently fails:
```Julia
Coloring: Test Failed at /home/runner/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterface/test/coloring.jl:11
  Expression: maximum(row_colors) < size(A, 1) ÷ 2
   Evaluated: 51 < 50

Stacktrace:
  [1] macro expansion
    @ /opt/hostedtoolcache/julia/1.10.2/x64/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
  [2] top-level scope
    @ ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterface/test/coloring.jl:508
```

gdalle:
It's a randomized test, should nearly never fail. Fix the seed and call it a day for now

DifferentiationInterface/docs/src/tutorial.md
```diff
@@ -6,45 +6,50 @@ CurrentModule = Main
...
 using DifferentiationInterface
-import ADTypes, ForwardDiff, Enzyme
-using BenchmarkTools
+
+import ForwardDiff, Enzyme  # ⚠️ import the backends you want to use ⚠️
 ```
 
-## Computing a gradient
+!!! tip
+    Importing backends with `import` instead of `using` avoids name conflicts and makes sure you are using operators from DifferentiationInterface.
```
gdalle:
```suggestion
    Importing backends with `import` instead of `using` avoids name conflicts and makes sure you are using operators from DifferentiationInterface.jl.
```

DifferentiationInterface/README.md
```diff
@@ -20,15 +20,15 @@ This package provides a backend-agnostic syntax to differentiate functions of th
...
+- First- and second-order operators (gradients, Jacobians, Hessians and [more](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/overview/))
 - In-place and out-of-place differentiation
 - Preparation mechanism (e.g. to create a config or tape)
 - Thorough validation on standard inputs and outputs (numbers, vectors, matrices)
 - Testing and benchmarking utilities accessible to users with [DifferentiationInterfaceTest](https://github.com/gdalle/DifferentiationInterface.jl/tree/main/DifferentiationInterfaceTest)
 
 ## Compatibility
 
-We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl) v1.0:
+We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl) `v1.0`:
```
gdalle:
```suggestion
We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl) v1.0:
```

DifferentiationInterface/README.md
```diff
@@ -68,19 +68,21 @@ julia> Pkg.add(
...
-julia> using DifferentiationInterface
+```julia
+using DifferentiationInterface
+import ForwardDiff, Enzyme, Zygote          # import automatic differentiation backends you want to use 
 
-julia> backend = ADTypes.AutoForwardDiff();
+f(x) = sum(abs2, x)
 
-julia> f(x) = sum(abs2, x);
+x = [1., 2., 3.]
```
gdalle:
```suggestion
x = [1.0, 2.0, 3.0]
```

gdalle (approved):

PR #218, opened 2024-04-26
Implement DifferentiateWith to translate between backends

gdalle:
**Core**

- Define `DifferentiateWith(f, backend)` to force differentiation with a given backend (exported but unstable API)
- Rename `backend_string` into `backend_str`

**Extensions**

- Add ChainRules for `DifferentiateWith` based on `value_and_pushforward` + `value_and_pullback_split`

**Tests**

- Improve structure of test set and progress logging (group by scenario type)
- Add utility for swapping the function inside a scenario
- Use that to define Zygote-breaking scenarios from the `default_scenarios()`
- Check that those scenarios don't error when we use `AutoZygote()` as an outer backend on `DifferentiateWith(f, AutoForwardDiff())`

**Documentation**

- Document `DifferentiateWith` (docstring with example + overview)

### DIFF
--- a/DifferentiationInterface/docs/src/api.md
+++ b/DifferentiationInterface/docs/src/api.md
@@ -91,6 +91,12 @@ check_twoarg
 check_hessian
 ```
 
+## Translation
+
+```@docs
+DifferentiateWith
+```
+
 ## Internals
 
 This is not part of the public API.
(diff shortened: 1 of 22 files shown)


adrhill:
This one is a hit!

PR #240, opened 2024-04-30
Update description of overview table

scheidan:
Clarify how to read (the very helpful!) the overview table.


gdalle:
Thanks for the suggestion but I'm not sure it's a good idea, cause if we make the column names any longer it probably becomes necessary to scroll to read the full table?

gdalle:
Can you try it locally to see how it looks? Push preview is not available for PRs from forks

gdalle (approved):
I just checked and it displays nicely, thanks for the contribution!

PR #241, opened 2024-05-01
Updates for Tapir

willtebbutt:
@gdalle as alluded to on slack, v0.2 of Tapir made breaking changes which affect DI. This PR updates provides updates which handle that.

I've not checked, but I'm assuming you're doing continuous delivery, so I've bumped the patch to make a release possible. Happy to change it if that's not what you're doing.


willtebbutt:
On a related note, Tapir now has a "safe mode", which I would like to force users to actively choose to turn on / off. Is this best done by modifying `ADTypes.AutoTapir` (in a non-breaking way), or should I add something in this package?

gdalle:
> I've not checked, but I'm assuming you're doing continuous delivery, so I've bumped the patch to make a release possible. Happy to change it if that's not what you're doing.

That's perfect, although we're already on 0.3.1 so it's gonna have to be 0.3.2

> On a related note, Tapir now has a "safe mode", which I would like to force users to actively choose to turn on / off. Is this best done by modifying ADTypes.AutoTapir (in a non-breaking way), or should I add something in this package?

Yeah this should go in ADTypes.jl. Is this information useful for dispatch (in which case it goes in a type parameter) or just at runtime (in which case a struct field will do)?

DifferentiationInterface/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.3.0"
+version = "0.3.1"
```
```suggestion
version = "0.3.2"
```

DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/onearg.jl
```diff
@@ -5,25 +5,23 @@ end
 
 function DI.prepare_pullback(f, ::AutoTapir, x, dy)
```
gdalle, line 6: We could use this opportunity to store a little more in the `extras`, typically the objects that we have to allocate on the fly in `value_and_pullback`: `dy_righttype` and `dx_righttype`.
Same remark in the two-argument version

gdalle (changes requested):
Thanks for the contribution!
Do you want to try and see if we can get allocation-free pullbacks by storing more stuff in the `extras` instead of creating new objects?

DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/onearg.jl
willtebbutt: If it's okay with you, I'd rather punt this for now. I don't have the time to figure out the best way to do this right now. Happy to revisit next week though.

gdalle: sounds good!

gdalle (approved):

gdalle:
This also asks the question of what to do when an AD package wants to use DI for testing, in terms of releasing breaking versions:

- You can't test with DI until I bump the compat
- I can't bump the compat unless you test with DI

willtebbutt:
Yeah, this is a good question. I actually don't know what the answer is -- maybe AD packages should maintain the DI extensions in their own repos? That way if they make breaking changes, they can continue to support the DI interface without needing to bump the DI version.

gdalle:
Yeah that is probably the right answer. AbstractDifferentiation.jl wanted to do that, but it never got enough traction. Let's wait a bit more while things are in flux, but in the medium term I could definitely see Tapir having a DI extension instead of the other way around

PR #254, opened 2024-05-12
Warm up for Tapir's rrule

gdalle:
Apply the `rrule` once during preparation for Tapir

@willtebbutt can you help me do the two-argument case?

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/twoarg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/twoarg.jl
@@ -2,10 +2,11 @@ struct TapirTwoArgPullbackExtras{R} <: PullbackExtras
     rrule::R
 end
 
-function DI.prepare_pullback(f!, y, ::AutoTapir, x, dy)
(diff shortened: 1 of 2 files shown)


DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/twoarg.jl
```diff
@@ -3,7 +3,9 @@ struct TapirTwoArgPullbackExtras{R} <: PullbackExtras
 end
 
 function DI.prepare_pullback(f!, y, ::AutoTapir, x, dy)
-    return TapirTwoArgPullbackExtras(build_rrule(f!, y, x))
+    rrule = build_rrule(f!, y, x)
+    # how to warm up here?
```
willtebbutt: Can you just call the method of `DI.value_and_pulback` defined a couple of lines below here?

gdalle: oh right, I was afraid it would create an infinite recursive loop but it won't

PR #259, opened 2024-05-14
Tapir Safe Mode

willtebbutt:
Follow up from recent ADTypes PR -- this PR ensures that if someone asks for safe mode in when constructing `AutoTapir`, they get safe mode.

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -43,7 +43,7 @@ DifferentiationInterfaceTrackerExt = "Tracker"
 DifferentiationInterfaceZygoteExt = "Zygote"
 
 [compat]
-ADTypes = "1.0.0"
+ADTypes = "1.2.0"
 ChainRulesCore = "1.23.0"
 Compat = "3,4"
 Diffractor = "=0.2.6"
--- a/DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/DifferentiationInterfaceTapirExt.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceTapirExt/DifferentiationInterfaceTapirExt.jl
@@ -18,7 +18,8 @@ using Tapir:
     NoRData,
     fdata,
     rdata,
-    __value_and_pullback!!
+    __value_and_pullback!!,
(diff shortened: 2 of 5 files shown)


willtebbutt:
Test failures because version 1.2 of ADTypes is not yet available.

gdalle:
The formatting check failed. Also you are polluting my test suite with `@info` messages, is there any way to get rid of them? Perhaps test the unsafe mode here, and both modes on your end?

willtebbutt:
> The formatting check failed. 

Hmm yes, that does appear to be what's happening. Have you considered deploying ReviewDog e.g. as done in [AbstractGPs](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/master/.github/workflows/format.yml) in order to make any formatting issues + fixes appear in the PR?

> Also you are polluting my test suite with @info messages, is there any way to get rid of them?

Hahaha, yes, I am doing that at present.

> Perhaps test the unsafe mode here, and both modes on your end?

Yeah, that's probably the right thing to do. Will fix these problems now.

gdalle:
> Have you considered deploying ReviewDog e.g. as done in [AbstractGPs](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/master/.github/workflows/format.yml) in order to make any formatting issues + fixes appear in the PR?

From what I've seen, these tools can be extremely noisy and make tons of comments on a single commit, so I like having that check as part of my test suite. Especially when I work on rather ambitious PRs, which is often the case in DI

willtebbutt:
> From what I've seen, these tools can be extremely noisy and make tons of comments on a single commit, so I like having that check as part of my test suite. Especially when I work on rather ambitious PRs, which is often the case in DI

That's completely reasonable.

willtebbutt:
Another bug on my end because testing whether something prints is hard. Will fix and then re-run the tests.

gdalle (approved):

gdalle:
The printing is not a problem for me, it's just a mild annoyance. Since the tests pass, can I merge and release this?

willtebbutt:
Yes, please do. The next patch version of Tapir will fix the printing issue anyway.

PR #266, opened 2024-05-17
Emphasize operator preparation in README

### LINKED ISSUE #256: README example
At present the main example in the README is:
```
import ForwardDiff, Enzyme, Zygote  # AD backends you want to use 

f(x) = sum(abs2, x)

x = [1.0, 2.0]

value_and_gradient(f, AutoForwardDiff(), x) # returns (5.0, [2.0, 4.0]) with ForwardDiff.jl
value_and_gradient(f, AutoEnzyme(),      x) # returns (5.0, [2.0, 4.0]) with Enzyme.jl
value_and_gradient(f, AutoZygote(),      x) # returns (5.0, [2.0, 4.0]) with Zygote.jl
```

Would it be better to make this example one which includes the `extras = prepare_gradient(f, backend, x)` etc stuff?

I'm just thinking about the issue we had on Tapir.jl the other day, where a user didn't realise that you should really be using `prepare_gradient` in general. My general point is: if a user sees one example, maybe they should see one which we think is likely to give best performance across all backends?

adrhill:
Closes #256.

~~Also removes syntax highlighting in favor of a jldoctest (@gdalle).~~

### DIFF
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -71,7 +71,7 @@ julia> Pkg.add(
 
 ## Example
 
-```jldoctest
+```julia
 using DifferentiationInterface
 import ForwardDiff, Enzyme, Zygote  # AD backends you want to use 
 
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -71,7 +71,7 @@ julia> Pkg.add(
 
 ## Example
 
-```julia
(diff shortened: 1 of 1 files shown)


adrhill:
Not sure testing our most high-level and stable API is worth the hassle and giving up syntax highlighting.

gdalle (approved):

PR #283, opened 2024-05-29
Fix broken link

roflmaostc:


gdalle (approved):
thanks!

adrhill:
I'm surprised Documenter didn't catch the broken URL, since the README is included in the docs. Can we fix this?

PR #291, opened 2024-05-31
Self-preparing inner closures for second order

### LINKED ISSUE #86: Preparation for second order
Constructing the right extras becomes very tricky when different inner/outer backends must be called in various ways on closure functions

gdalle:
Try to fix #86


gdalle:
Closing this because it violates a fundamental assumption of preparation: the function should not change its internal state

PR #303, opened 2024-06-05
Make Zygote extension depend on ForwardDiff too

gdalle:
**DI extensions**

- Add ForwardDiff to the list of triggers for the Zygote extension, since it is needed by the HVP


adrhill:
Does this mean Zygote can only be used if ForwardDiff is loaded as well?

gdalle:
Yes. In practice it doesn't change anything because ForwardDiff is a dependency of Zygote anyway, and dropping it would be a breaking Zygote change so we'd see it coming

PR #305, opened 2024-06-05
Implement `value_gradient_and_hessian`

### LINKED ISSUE #299: Return `gradient` with `hessian`, or `derivative` with `second_derivative`
As discussed in #296, defining the following operators would allow some backends to optimize the joint computation:

- `value_derivative_and_second_derivative`
- `value_gradient_and_hessian`

Main issue: with what kind of scenario do we test those?

gdalle:
**DI source**

- [x] Add `value_gradient_and_hessian` for dense and sparse backends
- [x] Adjust `extras` accordingly to prepare the gradient too
- [x] Add `maybe_inner`, `maybe_outer` and `maybe_dense_ad` logic to accommodate first-order backends

**DI extensions**

Add operator for:

- [x] FastDifferentiation
- [x] FiniteDiff
- [x] ForwardDiff
- [x] PolyesterForwardDiff
- [x] ReverseDiff
- [x] Symbolics
- [x] Zygote

**DI docs**

- [x] Document `value_gradient_and_hessian`

**DIT source**

- [x] Add correctness, benchmark and type stability tests for `value_gradient_and_hessian`

### DIFF
--- a/DifferentiationInterface/docs/src/api.md
+++ b/DifferentiationInterface/docs/src/api.md
@@ -93,6 +93,8 @@ hvp!
 prepare_hessian
 hessian
 hessian!
+value_gradient_and_hessian
+value_gradient_and_hessian!
 ```
 
 ## Utilities
--- a/DifferentiationInterface/docs/src/operators.md
+++ b/DifferentiationInterface/docs/src/operators.md
@@ -45,16 +45,16 @@ These operators are computed using the input `x` and a "seed" `v`, which lives e
 
 Several variants of each operator are defined.
 
-| out-of-place                | in-place                     | out-of-place + primal                            | in-place + primal                                |
-| :-------------------------- | :--------------------------- | :----------------------------------------------- | :----------------------------------------------- |
-| [`derivative`](@ref)        | [`derivative!`](@ref)        | [`value_and_derivative`](@ref)                   | [`value_and_derivative!`](@ref)                  |
+| out-of-place                | in-place                     | out-of-place + primal                            | in-place + primal                                 |
+| :-------------------------- | :--------------------------- | :----------------------------------------------- | :------------------------------------------------ |
+| [`derivative`](@ref)        | [`derivative!`](@ref)        | [`value_and_derivative`](@ref)                   | [`value_and_derivative!`](@ref)                   |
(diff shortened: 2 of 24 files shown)


gerlero:
@gdalle Thanks a lot for the work! I've already managed to test the new functionality with my projects and can confirm that it works.

Unfortunately for the case of `ForwardDiff` (although not a surprise to me given it's what prompted https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/122), the `DiffResults`-based implementation of `value_and_derivative` and `value_derivative_and_second_derivative` is a bit slower (by 5% in my test code that does a lot of derivation) than [`AbstractDifferentiation`'s implementation](https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/master/ext/AbstractDifferentiationForwardDiffExt.jl#L94) based on manipulating dual numbers.

```
with AbstractDifferentiation:
5.917 ms (1243 allocations: 111.25 KiB)

with DifferentiationInterface:
6.207 ms (1243 allocations: 111.25 KiB)
```

Although the difference is small, I can't justify the migration of my code from `AbstractDifferentiation` if it'll be slower. That is unless you're okay with also taking `AbstractDifferentiation`'s implementation verbatim here (even if it won't work for non-scalar functions)?

gdalle:
Wait, we can do `value_derivative_and_second_derivative` using the public ForwardDiff API and DiffResults? I thought this only worked for the `hessian` (which computes `gradient` as a side product if you supply a `DiffResult`).

In any case, yes, a PR would be welcome with a custom dual number implementation. Let's start with a version that works in the scalar case, and then make it work in the vector case.
Here is the file you need to modify: https://github.com/gdalle/DifferentiationInterface.jl/blob/main/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
And here are some utilities that are useful for writing code that works on both scalar and vector output: https://github.com/gdalle/DifferentiationInterface.jl/blob/main/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/utils.jl
You can take a look at `DI.pushforward` to get inspiration.

gdalle:
Do you want to open said PR? I don't have complete confidence in my ability to get your code working on vectors, so it definitely would be great if you could help.

gerlero:
Sure, I'll start it later today (also don't think I can get it working on vectors, but it's worth a try)

Issue #134, opened 2024-04-03
Possible direct users


gdalle:
Meta-issue to survey users of DifferentiationInterface and the related discussions.

### Registered

**[ContinuousNormalizingFlows.jl](https://github.com/impICNF/ContinuousNormalizingFlows.jl)**

- https://github.com/impICNF/ContinuousNormalizingFlows.jl/issues/397
- https://github.com/impICNF/ContinuousNormalizingFlows.jl/pull/399

**[CTBase.jl](https://github.com/control-toolbox/CTBase.jl) and [CTFlows.jl](https://github.com/control-toolbox/CTFlows.jl)**

- https://github.com/control-toolbox/CTBase.jl/issues/25
- https://github.com/control-toolbox/CTBase.jl/pull/140
- https://github.com/control-toolbox/CTBase.jl/pull/141

**[Fronts.jl](https://github.com/gerlero/Fronts.jl)**

- https://github.com/gerlero/Fronts.jl/pull/184

**[ImplicitDifferentiation.jl](https://github.com/JuliaDecisionFocusedLearning/ImplicitDifferentiation.jl)**

- https://github.com/JuliaDecisionFocusedLearning/ImplicitDifferentiation.jl/pull/135

**[ItemResponseFunctions.jl](https://github.com/JuliaPsychometrics/ItemResponseFunctions.jl)**

- https://github.com/JuliaPsychometrics/ItemResponseFunctions.jl/pull/14

**[MarginalLogDensities.jl](https://github.com/ElOceanografo/MarginalLogDensities.jl)**

- https://github.com/ElOceanografo/MarginalLogDensities.jl/issues/13
- https://github.com/ElOceanografo/MarginalLogDensities.jl/pull/25

**[OptimizationBase.jl](https://github.com/SciML/OptimizationBase.jl) and [Optimization.jl](https://github.com/SciML/Optimization.jl)**

- https://github.com/gdalle/DifferentiationInterface.jl/issues/89
- https://github.com/SciML/OptimizationBase.jl/pull/31
- https://github.com/SciML/Optimization.jl/issues/744
- https://github.com/SciML/OptimizationBase.jl/issues/62
- https://github.com/SciML/OptimizationBase.jl/pull/54

**[SimpleNonlinearSolve.jl](https://github.com/SciML/SimpleNonlinearSolve.jl)**

- https://github.com/SciML/SimpleNonlinearSolve.jl/pull/148

### Not registered

[...]

PR #310, opened 2024-06-06
Add dual number–based second derivatives for `ForwardDiff`

gerlero:
As discussed in #305, I'm starting by copying the implementations from https://github.com/JuliaDiff/AbstractDifferentiation.jl/pull/122 almost verbatim.

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/DifferentiationInterfaceForwardDiffExt.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/DifferentiationInterfaceForwardDiffExt.jl
@@ -10,6 +10,7 @@ using DifferentiationInterface:
     HessianExtras,
     JacobianExtras,
     NoDerivativeExtras,
+    NoSecondDerivativeExtras,
     PushforwardExtras
 using ForwardDiff.DiffResults: DiffResults, DiffResult, GradientResult, MutableDiffResult
 using ForwardDiff:
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
@@ -68,7 +68,7 @@ function DI.prepare_second_derivative(f::F, backend::AutoForwardDiff, x) where {
 end
 
 function DI.second_derivative(
-    f::F, ::AutoForwardDiff, x, dx, ::NoSecondDerivativeExtras
+    f::F, ::AutoForwardDiff, x, ::NoSecondDerivativeExtras
 ) where {F}
     T = tag_type(f, backend, x)
     xdual = make_dual(T, x, one(x))
@@ -78,7 +78,7 @@ function DI.second_derivative(
 end
 
 function DI.second_derivative!(
-    f::F, der2, ::AutoForwardDiff, x, dx, ::NoSecondDerivativeExtras
+    f::F, der2, ::AutoForwardDiff, x, ::NoSecondDerivativeExtras
 ) where {F}
     T = tag_type(f, backend, x)
     xdual = make_dual(T, x, one(x))
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
@@ -67,14 +67,6 @@ function DI.prepare_derivative(f::F, backend::AutoForwardDiff, x) where {F}
     return NoDerivativeExtras()
 end
 
-function DI.derivative(f::F, ::AutoForwardDiff, x, ::NoDerivativeExtras) where {F}
-    return derivative(f, x)
-end
-
-function DI.derivative!(f::F, der, ::AutoForwardDiff, x, ::NoDerivativeExtras) where {F}
-    return derivative!(der, f, x)
-end
-
 function DI.value_and_derivative(
     f::F, backend::AutoForwardDiff, x, ::NoDerivativeExtras
 ) where {F}
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
@@ -71,12 +71,25 @@ function DI.derivative(f::F, ::AutoForwardDiff, x, ::NoDerivativeExtras) where {
     return derivative(f, x)
 end
 
+function DI.derivative!(f::F, der, ::AutoForwardDiff, x, ::NoDerivativeExtras) where {F}
+    return derivative!(der, f, x)
+end
+
 function DI.value_and_derivative(
     f::F, backend::AutoForwardDiff, x, ::NoDerivativeExtras
 ) where {F}
     T = tag_type(f, backend, x)
     ydual = f(make_dual(T, x, one(x)))
     return myvalue(T, ydual), myderivative(T, ydual)
+end
+
+function DI.value_and_derivative!(
+    f::F, der, ::AutoForwardDiff, x, ::NoDerivativeExtras
+) where {F}
+    T = tag_type(f, backend, x)
+    xdual = make_dual(T, x, one(x))
+    ydual = f(xdual)
+    return myvalue(T, ydual), myderivative!(T, der, ydual)
 end
 
 function DI.value_derivative_and_second_derivative(
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
@@ -63,6 +63,10 @@ end
 
 ## Derivative
 
+function DI.prepare_derivative(f::F, backend::AutoForwardDiff, x) where {F}
+    return NoDerivativeExtras()
+end
+
 function DI.value_and_derivative(f::F, backend::AutoForwardDiff, x) where {F}
     T = tag_type(f, backend, x)
     ydual = f(make_dual(T, x, one(x)))
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
@@ -61,6 +61,27 @@ function DI.pushforward!(
     return dy
 end
 
+## Derivative
+
+function DI.value_and_derivative(f::F, backend::AutoForwardDiff, x) where {F}
+    T = tag_type(f, backend, x)
+    ydual = f(make_dual(T, x, one(x)))
+    return myvalue(T, ydual), myderivative(T, ydual)
+end
+
+function DI.value_derivative_and_second_derivative(
+    f::F, backend::AutoForwardDiff, x
+) where {F}
+    T = tag_type(f, backend, x)
+    xdual = make_dual(T, x, one(x))
+    T2 = tag_type(f, backend, xdual)
+    ydual = f(make_dual(T2, xdual, one(xdual)))
+    v = myvalue(T, myvalue(T2, ydual))
+    d = myderivative(T, myvalue(T2, ydual))
+    d2 = myderivative(T, myderivative(T2, ydual))
+    return v, d, d2
+end
+
 ## Gradient
 
 struct ForwardDiffGradientExtras{C} <: GradientExtras
@@ -182,23 +203,4 @@ function DI.value_gradient_and_hessian(
     return (
         DiffResults.value(result), DiffResults.gradient(result), DiffResults.hessian(result)
     )
-end
-
-function DI.value_and_derivative(f::F, backend::AutoForwardDiff, x) where {F}
-    T = tag_type(f, backend, x)
-    ydual = f(make_dual(T, x, one(x)))
-    return myvalue(T, ydual), myderivative(T, ydual)
-end
-
-function DI.value_derivative_and_second_derivative(
-    f::F, backend::AutoForwardDiff, x
-) where {F}
-    T = tag_type(f, backend, x)
-    xdual = make_dual(T, x, one(x))
-    T2 = tag_type(f, backend, xdual)
-    ydual = f(make_dual(T2, xdual, one(xdual)))
-    v = myvalue(T, myvalue(T2, ydual))
-    d = myderivative(T, myvalue(T2, ydual))
-    d2 = myderivative(T, myderivative(T2, ydual))
-    return v, d, d2
 end
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl
@@ -184,19 +184,21 @@ function DI.value_gradient_and_hessian(
     )
 end
 
-function DI.value_and_derivative(f::F, ::AutoForwardDiff, x) where {F}
-    T = typeof(Tag(f, typeof(x)))
-    ydual = f(Dual{T}(x, one(x)))
-    return value(T, ydual), extract_derivative(T, ydual)
+function DI.value_and_derivative(f::F, backend::AutoForwardDiff, x) where {F}
+    T = tag_type(f, backend, x)
+    ydual = f(make_dual(T, x, one(x)))
+    return myvalue(T, ydual), myderivative(T, ydual)
 end
 
-function DI.value_derivative_and_second_derivative(f::F, ::AutoForwardDiff, x) where {F}
-    T = typeof(Tag(f, typeof(x)))
-    xdual = Dual{T}(x, one(x))
-    T2 = typeof(Tag(f, typeof(xdual)))
-    ydual = f(Dual{T2}(xdual, one(xdual)))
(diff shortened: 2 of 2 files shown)


gdalle:
You can disregard the Enzyme failures, they're obviously unrelated

gdalle:
Can you try using the functions in the `utils.jl` file to make this scalar-/vector- agnostic?

gerlero:
> Can you try using the functions in the `utils.jl` file to make this scalar-/vector- agnostic?

I'll give it a try

gdalle:
Cool, thanks! It should be similar to the pushforward one, albeit more intricate

gerlero:
I'm using the functions from `utils.jl` now, but the code still won't work with vector inputs due to my use of `make_dual(T, x, one(x))` (as `one(x)` is not defined for vectors).

gerlero:
It seems my last comment was due to some confusion. In [a previous comment](https://github.com/gdalle/DifferentiationInterface.jl/discussions/296#discussioncomment-9658367) you only mention the need to handle $\mathbb{R}\to\mathbb{R}^n$ functions, so vector inputs should not expected to work with the `derivative` functions. Is this correct?

gerlero:
After a small fixup, the new functionality should be working now. I've tested it and it seems to handle vector-valued (w/ scalar input) functions just fine.

gdalle:
Yes, derivative is for scalar to scalar or scalar to vector. Thank you for the contribution, I'll allow the workflows and review tomorrow

gdalle:
Your code is never touched because you did not overload preparation, or provide methods which accept the additional "extras" argument. I'll fix it tomorrow, but if you want to do it yourself, you can take inspiration from the other methods in the extension

gerlero:
>  I'll fix it tomorrow, but if you want to do it yourself, you can take inspiration from the other methods in the extension

I tried it but had to give up (at least for now), as I'm not familiar at all with how extras and `pushforward` interact with each other (I tried returning a `NoDerivativeExtras` from `prepare_derivative`, but then I get ``ERROR: type NoDerivativeExtras has no field pushforward_extras`` which is not obvious to me how to fix). I'm guessing you have a much more solid idea in your mind for this part.

gdalle:
Indeed, the current version of the package gives errors like this one, which are very hard to parse:

```julia
type NoDerivativeExtras has no field pushforward_extras
  Stacktrace:
    [1] getproperty(x::DifferentiationInterface.NoDerivativeExtras, f::Symbol)
      @ Base ./Base.jl:37
    [2] value_and_derivative!(f::DifferentiationInterfaceTest.var"#num_to_arr#13"{StaticArraysCore.SMatrix{2, 3, Int64, 6}}, der::StaticArraysCore.MMatrix{2, 3, Float64, 6}, backend::AutoForwardDiff{2, Symbol}, x::Float64, extras::DifferentiationInterface.NoDerivativeExtras)
      @ DifferentiationInterface ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterface/src/first_order/derivative.jl:91
```

I thus coded #313 which should make errors much more pleasant, let's see how it looks after merging `main` into the current branch

gdalle:
And the new error is much more informative! I'll add these indications to the docs, but essentially it tells you that you forgot to implement some necessary methods:

```julia
MethodError: no method matching derivative(::DifferentiationInterfaceTest.var"#num_to_arr#13"{StaticArraysCore.SMatrix{2, 3, Int64, 6}}, ::AutoForwardDiff{2, Symbol}, ::Float64, ::DifferentiationInterface.NoDerivativeExtras)
  
  Closest candidates are:
    derivative(::F, ::Any, ::AutoForwardDiff, ::Any, ::DifferentiationInterfaceForwardDiffExt.ForwardDiffTwoArgDerivativeExtras) where F
     @ DifferentiationInterfaceForwardDiffExt ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/twoarg.jl:95
    derivative(::F, ::Any, ::ADTypes.AbstractADType, ::Any, ::DifferentiationInterface.PushforwardDerivativeExtras) where F
     @ DifferentiationInterface ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterface/src/first_order/derivative.jl:147
    derivative(::F, ::Any, ::ADTypes.AbstractADType, ::Any) where F
     @ DifferentiationInterface ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterface/src/first_order/derivative.jl:127
    ...
```

Whenever you implement a new `operator`, you need to provide

- the preparation step `prepare_operator`
- the four variants `operator`, `operator!`, `value_and_operator`, `value_and_operator!`

Sorry to use you as a bit of a lab rat, but you're the first one to try and contribute something nontrivial to DI (for which I'm grateful!). That means it's a good test for whether our process is clear enough or not (obviously not)

gerlero:
> Whenever you implement a new `operator`, you need to provide
> 
> * the preparation step `prepare_operator`
> * the four variants `operator`, `operator!`, `value_and_operator`, `value_and_operator!`
> 
> Sorry to use you as a bit of a lab rat, but you're the first one to try and contribute something nontrivial to DI (for which I'm grateful!). That means it's a good test for whether our process is clear enough or not (obviously not)

No problem. Thanks! I've added `derivative` now too.

Not sure about `derivative!` and `value_and_derivative!`, though: are those supposed to return `ForwardDiff.DiffResults` structs or some other DI-specific format?

gdalle:
You can ask the docstrings: we return a tuple of the value and its derivative, whatever type they have, so no DiffResults.

gdalle:
Do you want me to finish the PR? I definitely can if you get tired of the DI kinks. But having one more autodiff-interested individual look at the guts of the code is rather instructive for me

gerlero:
I'll make my attempt to finish it

gerlero:
All right, I've added the missing methods for `derivative!` and `value_and_derivative!` (not sure if there's anything left to do).

gdalle:
I know what's going on!
Sorry for not clocking it earlier, but I think you're targeting the wrong operator here. You noticed that `DI.derivative` wasn't implemented in the ForwardDiff extension for one-argument functions. The reason is that it the derivative exactly a pushforward, and the existing pushforward code is basically optimal there.
The operator you're interested in is `second_derivative`, and if you look at the table [here](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/operators/#Variants), you'll find its 4 variants are the following ones:
- `second_derivative`
- `second_derivative!`
- `value_derivative_and_second_derivative`
- `value_derivative_and_second_derivative!`

So you actually need to code these, plus a suitable `prepare_second_derivative`.
Again, if this is no fun for you, I can take it off your hands anytime.

gdalle:
(in addition there are test failures due to typos but that's an easy fix)

gdalle:
You can also drop `value_and_derivative` and its in-place counterpart. This too is taken care of by `value_and_pushforward` in an essentially optimal way. Your focus here is only the second derivative, and its `value_derivative_and_second_derivative` friend

gdalle:
This will inspire lots of developer doc fixes for me, thank you so much for putting up with me. Essentially, DI thinks of the four variants (in-place or out-of-place, with or without `value`) as "the same operator", wihch is why they always go together

gerlero:
> You can also drop `value_and_derivative` and its in-place counterpart. This too is taken care of by `value_and_pushforward` in an essentially optimal way.

Are we sure that doesn't rely on DiffResults at all?

gdalle:
> Are we sure that doesn't rely on DiffResults at all?

- `derivative(f!, y, backend, x)` (with its variants) relies on the ForwardDiff + DiffResults API, see [here](https://juliadiff.org/ForwardDiff.jl/stable/user/api/#Derivatives-of-f(x::Real)::Union{Real,AbstractArray})
- `derivative(f, backend, x)` (with its variants) has no equivalent API to rely on, at least not a public one, which is why I don't think we gain much by going beyond what already exists for the pushforward?

gerlero:
> * `derivative(f, backend, x)` (with its variants) has no equivalent API to rely on, at least not a public one, which is why I don't think we gain much by going beyond what already exists for the pushforward?

Great. I'm still not sure what the fallback is for each method so I just wanted to check (a point of having `value_and_derivative`/`value_derivative_and_second_derivative` AFAIC is to avoid having to use the slower DiffResults API).

gdalle:
Whenever it exists, I prefer sticking to the backend's existing API, in particular the public parts. My rationale is that I can't afford to try and optimize 13 backends better than their own individual creators, so I make do with what said creators give me. ForwardDiff might be a bit of an exception because technically the `Dual` parts of the package are not public, but you get the general idea. If DiffResults is slow, it's on them, not on me.

If you want to investigate how `value_and_derivative(f, backend, x)` works for ForwardDiff without an explicit implementation, it falls back on `pushforward` via the following call chain:

**Generic**

https://github.com/gdalle/DifferentiationInterface.jl/blob/a99715453fe269f5a8f00d66d1a81fe3480acbd3/DifferentiationInterface/src/first_order/derivative.jl#L58-L65

https://github.com/gdalle/DifferentiationInterface.jl/blob/a99715453fe269f5a8f00d66d1a81fe3480acbd3/DifferentiationInterface/src/first_order/derivative.jl#L75-L77

https://github.com/gdalle/DifferentiationInterface.jl/blob/a99715453fe269f5a8f00d66d1a81fe3480acbd3/DifferentiationInterface/src/first_order/derivative.jl#L91-L95

**Specific**

https://github.com/gdalle/DifferentiationInterface.jl/blob/a99715453fe269f5a8f00d66d1a81fe3480acbd3/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl#L30-L37

https://github.com/gdalle/DifferentiationInterface.jl/blob/a99715453fe269f5a8f00d66d1a81fe3480acbd3/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl#L3-L19

Do you think we can get better than this?

gerlero:
> Whenever it exists, I prefer sticking to the backend's existing API, in particular the public parts. My rationale is that I can't afford to try and optimize 13 backends better than their own individual creators, so I make do with what said creators give me. ForwardDiff might be a bit of an exception because technically the `Dual` parts of the package are not public, but you get the general idea. If DiffResults is slow, it's on them, not on me.

Yes, I know, and appreciate the exception you're making here! In fact, I have an open PR (https://github.com/JuliaDiff/ForwardDiff.jl/pull/678) at ForwardDiff asking for this to be implemented there, but with no luck so far.

> If you want to investigate how `value_and_derivative(f, backend, x)` works for ForwardDiff without an explicit implementation, it falls back on `pushforward` via the following call chain:

> Do you think we can get better than this?

Reading through the code, I don't think so. Also, in my benchmark case this PR is now running as fast as `AbstractDifferentiation`.

gdalle:
Spoiler alert, I think (hope) the tests will fail because you didn't implement `second_derivative` or `second_derivative!`

gerlero:
> Spoiler alert, I think (hope) the tests will fail because you didn't implement `second_derivative` or `second_derivative!`

Ouch! I'll add them.

gdalle:
Is it just me or are the checks running forever?

gerlero:
Sorry, I had an extra parameter (not sure where it came from) in second_derivative(!). Let's see now.

I'm also now running the tests locally to see what happens.

gdalle:
This file should be enough. Don't run the whole test suite you will need an hour or so

https://github.com/gdalle/DifferentiationInterface.jl/blob/main/DifferentiationInterfaceTest/test/forwarddiff.jl

gerlero:
> This file should be enough. Don't run the whole test suite you will need an hour or so
> 
> https://github.com/gdalle/DifferentiationInterface.jl/blob/main/DifferentiationInterfaceTest/test/forwarddiff.jl

That's what I did. Caught another bug :/. Running again to see if it's fixed

gdalle:
When the tests fail, do you usually find it clear what you need to fix?

gerlero:
Should be fixed now. I'm getting a couple of efficiency test errors with `ForwardDiff`, but I'm not sure they are related.

> When the tests fail, do you usually find it clear what you need to fix?

Yes, you get the stack trace right there as soon as a test fails, which is nice. IDK how it could be made better.

gdalle (approved):
Thanks for the contribution!

gerlero:
Thank you! Can I ask for a new release with this change?

gdalle:
Yes, i was just waiting to check that the tests pass on main.
Did you figure out why the efficiency tests were failing? Are you on a different Julia version perhaps?

gerlero:
> Did you figure out why the efficiency tests were failing? Are you on a different Julia version perhaps?

Not really; I just saw that something is allocating when it shouldn't. FWIW these are the errors:

```
...
[ Info: Testing Single/ForwardDiff/efficiency.jl
derivative! - vecexp! : Float64 -> Vector{Float64}: Test Failed at /Users/gabriel/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:49dDiff (forward) - 1/1
  Expression: row[:allocs] == 0rio - 2/2
   Evaluated: 1.0 == 0
  arguments:      2
Stacktrace:       inplace
 [1] macro expansioncexp!
   @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion0,)
   @ ~/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:49 [inlined]
 [3] macro expansion0,)
   @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] macro expansion
   @ ~/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:49 [inlined]
 [5] macro expansion
   @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [6] top-level scope
   @ ~/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:18
jacobian! - vecexp! : Vector{Float64} -> Vector{Float64}: Test Failed at /Users/gabriel/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:49
  Expression: row[:allocs] == 0
   Evaluated: 1.0 == 0

Stacktrace:
 [1] macro expansion
   @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:49 [inlined]
 [3] macro expansion
   @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] macro expansion
   @ ~/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:49 [inlined]
 [5] macro expansion
   @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [6] top-level scope
   @ ~/DifferentiationInterface.jl/DifferentiationInterface/test/Single/ForwardDiff/efficiency.jl:18
[ Info: Testing Single/ForwardDiff/test.jl
...
```

I'm on 1.10.4 on macOS arm64.

gdalle:
Very strange!
Anyway, the registration is triggered, should be merged in 15 minutes: https://github.com/JuliaRegistries/General/pull/108479
Can I ask what packages you might wanna use DI for? Keeping this issue updated brings me great joy: https://github.com/gdalle/DifferentiationInterface.jl/issues/134

gerlero:
> Can I ask what packages you might wanna use DI for? Keeping this issue updated brings me great joy: https://github.com/gdalle/DifferentiationInterface.jl/issues/134

I'll be using it in [Fronts](https://github.com/gerlero/Fronts.jl) to replace `AbstractDifferentiation` as it is used in [this file](https://github.com/gerlero/Fronts.jl/blob/main/src/odes.jl).

I should be ready to migrate as soon as the new release is live. I'll make sure to mention the relevant PR in #134.

gdalle:
Interesting! Since it will also allow you to try other backends, I wonder if symbolic backends might be close to optimal for such a one-dimensional case. What kind of functions do you often end up differentiating?

gerlero:
> Interesting! Since it will also allow you to try other backends, I wonder if symbolic backends might be close to optimal for such a one-dimensional case.

Thanks. It's probably worth a try. I should first learn how symbolic differentiation works with DI, but I think it's worth testing to see if I can use it to shave a few milliseconds off common cases.

> What kind of functions do you often end up differentiating?

Functions like this one (plus others also present in the `PorousModels` directory):

https://github.com/gerlero/Fronts.jl/blob/78bea2f1f6f2f363a461f2c3c7a2e912fd5ea072/src/PorousModels/letd.jl#L45

gdalle:
For symbolic differentiation, preparation (the extras shenanigans) is very costly but absolutely essential, since it compiles executable versions of your function and its derivatives. At the moment I haven't super optimized that part because hardly anyone uses it. But if you want to use it, I'll definitely squeeze out some more performance

gerlero:
Thanks. I'll give it a try.

Any recommendation for which symbolic backend I can start with? `FastDifferentiation`?

PR #329, opened 2024-06-24
Add the GTPSA.jl backend

### LINKED ISSUE #316: Adding `GTPSA.jl` to the interface
I'll use this issue for questions I have for implementing this interface into [`GTPSA.jl`](https://github.com/bmad-sim/GTPSA.jl). I've done some reading of the documentation and those AD packages in `ext`, and here are some of my first questions

1. As long as I also overload `value_and_gradient`, `value_and_jacobian`, etc, is it safe to assume `pushforward`, `value_and_pushforward` only receive single variable functions?
2. Should I basically plan to overload all of the functions [here](https://docs.juliahub.com/General/DifferentiationInterface/stable/api/)?

mattsignorelli:
#316


gdalle:
@mattsignorelli thank you for the contribution!
I fixed a few things and defined `AutoGTPSA` in the main package for now. Types should not be defined in package extensions because it makes them hard to import.
The main part missing is the pushforward operator. If you don't have a dedicated mechanism for that, you can always deduce it from `derivative`, `gradient` or `jacobian` depending on the function type (array/scalar -> array/scalar). It will be slow but it will work.

mattsignorelli:
Sorry for the delay on this, I've had my hands tied with all different projects. Two weeks ago I pushed major updates/fixes and finishing touches to `GTPSA.jl`, which is now it's in first stable release. I'll get back to this shortly

gdalle:
The structure of the tests has changed a little since the last time you checked, can you merge `main` into this PR?

mattsignorelli:
OK everything looks good. I had to release a new GTPSA (v1.1) to allow its jacobian to accept `AbstractArray`, and also define the `broadcastable` trait. But `Back/GTPSA` now passes all `test_differentiation` tests. I commented out the coverage tests in the yml because it was erroring due to some token thing I do not have set up on my own fork

mattsignorelli:
Sorry accidentally hit the close button instead of submit... time for bed

mattsignorelli:
I submitted a PR to ADTypes here https://github.com/SciML/ADTypes.jl/pull/78

gdalle:
Let me just take the time to review here, I haven't yet looked at your code. If all is well on the DI front, we can add `AutoGTPSA` to ADTypes!
Thank you so much for your contribution

gdalle:
@mattsignorelli before I review, I updated the tests so that we now have full code coverage information. Can you take a look and see if you are able to increase the coverage?

One thing you need to know is that in-place operators like `pushforward!(f, dy, backend, x, dx)` are not tested on scenarios for which the target `dy` is not mutable. So the branch `if y isa Number` in this operator is not needed.

Another thing you may want to exploit to reduce code duplication is the redundancy between operators: `derivative` is just `pushforward` with `dx = 1`, `gradient` is just `pullback` with `dy = 1`. They all fall back onto one another inside DI, so that it is theoretically sufficient to just implement `value_and_pushforward`. If custom implementations of other operators yield a speed boost, then let's keep them, but if all you do in `derivative` is essentially call `pushforward`, you can get rid of it (for example).

mattsignorelli:
Thanks for looking this over!

For second order, because GTPSA does not allow nesting, separate operators must be written. For the first order operators `gradient`, `jacobian`, there is a large slowdown/significantly higher number of allocations when using only `pushforward`. This is because, in order to construct the `jacobian` or `gradient`, several `pushforward`s with different seeds have to be propagated through the function. The `TPS` type is mutable, and so it is much slower to repeatedly evaluate the function for different seeds, than it is to evaluate it once treating the `TPS`s as variables and then extract the first order derivatives. However `derivative` I have removed as using `pushforward` for that is equivalent.

I have also removed inaccessible branches (where `y` is not mutable) for the in-place operators, following the code coverage report, and simplified the HVP code (since the best GTPSA can do really is construct the enter hessian and multiply the vector).

The rest of the uncovered code is because the user can optionally specify a custom `Descriptor`, which lives in the `AutoGTPSA` type - this allows users to choose custom truncation orders for different variables (e.g. calculate hessian of `f(x,y)` but assume no $x^2$ terms in the resulting truncated power series so $\partial^2 f/\partial x^2$ is zero). This can significantly improve speed at the cost of lower accuracy. This feature is only applicable in highly specific use cases and I don't expect it to be used very often, but it should be available since GTPSA provides it. As for the code coverage in this case, the number of input variables and truncation order(s) have to be specified in the `Descriptor` for each tested function, and I'm not sure how to do this for every test case?

gdalle:
Sorry for taking so long to get back to you @mattsignorelli! I have brought your PR up to speed with the rest of the package, and taken into account the new version of ADTypes which includes `AutoGTPSA`. 
The last thing I'd like to do before merging is clean up the code a little bit, by removing duplication in favor of clear auxiliary functions. For example, my last commit cuts a lot of code from `pushforward` thanks to
```julia
function initialize!(xt::TPS, x::Number, dx::Number)
    xt[0] = x
    xt[1] = dx
    return xt
end

function initialize!(xt::AbstractArray{<:TPS}, x::AbstractArray, dx::AbstractArray)
    for i in eachindex(xt, x, dx)
        initialize!(xt[i], x[i], dx[i])
    end
    return xt
end
```
Can you maybe create similar functions to shorten the other operators? Just like `TPS` initialization, extraction from `TPS` into numbers and arrays could also be streamlined.

mattsignorelli:
Thanks for your work on this!

> ```julia
> function initialize!(xt::TPS, x::Number, dx::Number)
>     xt[0] = x
>     xt[1] = dx
>     return xt
> end
> ```

I think we need to be careful here. This would only apply for single-variable functions: `xt[1] = dx` basically sets the dual part of the first variable to `dx`. However, if there are two variables, then to set the dual part of the second variable, you would need to do `xt[2] = dx`.  E.g.

```julia
julia> d = Descriptor(2,1); # two variables to first order

julia> f = TPS(use=d);

julia> f[0] = 1; f[1] = 2; f[2] = 3;

julia> print(f) # corresponds to f(dx1,dx2) = 1 + dx1 + dx2
TPS64:
 Coefficient                Order   Exponent
  1.0000000000000000e+00      0      0   0
  2.0000000000000000e+00      1      1   0
  3.0000000000000000e+00      1      0   1
```

Also, because the `TPS` type is mutable, we really only want to evaluate the function once. I'll make some changes to the updates and try to simplify/remove code duplication using utility functions, and push for review

mattsignorelli:
I think I understand what `Tangents` are now. Just to check: to take the pushforward of a 3 variable function with seed `[dx1, dx2, dx3]`, your `Tangents` object will be `Tangents{1, Vector{Float64}}(([dx1, dx2, dx3]))` ? And so `Tangents` is there to store multiple seeds?

gdalle:
> Tangents is there to store multiple seeds?

The expression "multiple seeds" is a bit unclear. For instance, for a function `f(x::AbstractVector)`, DI considers `dx::AbstractVector` to be "a single seed". It used to be possible to call `pushforward(f, backend, x, dx)` but that is no longer the case.
From v0.6 you will need to use the wrapper `pushforward(f, backend, x, Tangents(dx))`, and recover a `Tangents{1}` as the result. On the other hand, if you have several seeds `dxa`, `dxb`, `dxc` (perhaps clearer than numbers?) inside `Tangents(dxa, dxb, dxc)`, then your batch size is 3 and you will get a `Tangents{3}` as the result.
Sorry if this is not well documented, you're the first outside person to ever contribute an extension so I didn't plan for that case ^^ But if there are unclear aspects I should definitely improve the docs.

gdalle:
> I think we need to be careful here. This would only apply for single-variable functions

That's the setting considered in DI. Even with all the latest changes, there will only ever be one active variable `x`, either in `f(x)` or in `f!(y, x)`. Of course that variable can itself be an array.

mattsignorelli:
Ah ok, I think I understand better. I'm admittedly not super familiar with the usual conventions used by most AD packages in Julia, only that in GTPSA, so I'm trying to connect the dots here. Let me explain how it works in GTPSA, and then maybe you can help better see how to fit it into DI: 

Basically with GTPSA, it calculates truncated power series in the variables, assuming the variables are small. So for example to calculate a Jacobian of a function $F: \\mathbb{R}^n -> \mathbb{R}^m$, you would define a `Descriptor(n, 1)` for $n$ variables to order 1, and then use `x = vars(d)` which gives you a vector which in the usual AD language is actually $[\Delta x_1, \Delta x_2, ... \Delta x_n]$ ("variables" in GTPSA are small, so actually this might be better understood as `dx = vars(d)`). Then, to calculate the multivariable Taylor expansion of the function $F$ around some point $\vec{a}$, you would just do `y = F(a.+x)`. `y` here is now the multivariable Taylor expansion. You can then extract the Jacobian from the monomial coefficients of the Taylor expansion.

Alternatively, one could do this one with variable, e.g. calculate $F([a_1+\Delta x_1, a_2, ..., a_n])$, then do $F([a_1, a_2+\Delta x_1, ..., a_n])$ etc and extract the first order part of the `TPS` result for each component for each calculation. However, this will be slower than instead propagating all of  $F(\vec{a}+[\Delta x_1, \Delta x_2, ... \Delta x_n])$, because `TPS`s are mutable and we only want to calculate the function once.

gdalle:
That makes sense, thank you. But then I think the `initialize!` function I wrote earlier is fine? At least it passes the tests ^^
Also, this makes me wonder if we could compute HVPs by considering both `x` and `dx` as variables somehow. With higher-level differentiation utilities like those in GTPSA, there must be something better to do than computing the full Hessian first.

mattsignorelli:
OK sorry for the big delay, I have been completely consumed by my other project but I'm ready to get back to this now.

> Also, this makes me wonder if we could compute HVPs by considering both `x` and `dx` as variables somehow. With higher-level differentiation utilities like those in GTPSA, there must be something better to do than computing the full Hessian first.

That's a great idea. I'll look into this and see if I can come up with anything.

I'll take a look at the latest code and dev guide to reorient myself, and be in touch soon.

mattsignorelli:
@gdalle This should be ready soon.. For HVP, I'm having difficulty understanding the tangents. For example one of the tests gives a 6x6 Hessian matrix but the tangent is a scalar number. Shouldn't the tangent be a vector?

mattsignorelli:
OK other than `hvp`, we are pretty much done, all other tests pass. In order to achieve full code coverage, I'm adding the kwarg `unsafe_inbounds` to GTPSA's getter functions, so that I can run tests with `AutoGTPSA{Descriptor}`, and just let the `Descriptor` have many variables to second order (before was just running tests with `AutoGTPSA{Nothing}`). Once this is released I will bump the version. 

I'm having a bit of trouble with `hvp`. It's basically the same code that worked some number of months ago, but now it's showing disagreements. Any help or clarification on my comment above would be much appreciated :) On that note, I looked into whether or not there is a faster way to calculate the hvp without materializing the full hessian, and unfortunately because `TPS`s cannot be nested, there isn't really any other option than materializing the full hessian. There are some other ways I found, but none of them are faster..

mattsignorelli:
Actually, I realized I made a mistake in which tests are being run when I made my previous comment. I found the problem

mattsignorelli:
This is ready for merge now

gdalle:
thanks, I'll take a look!

mattsignorelli:
Hi any updates on this?

DifferentiationInterface/docs/Project.toml
```diff
@@ -7,6 +7,7 @@ DocumenterInterLinks = "d12716ef-a0f6-4df4-a9f1-a5a34e75c656"
 DocumenterMermaid = "a078cd44-4d9c-4618-b545-3ab9d77f9177"
 FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+GTPSA = "b27dd330-f138-47c5-815b-40db9dd9b6e8"
```
gdalle: It is actually no longer needed here
```suggestion
```

DifferentiationInterface/docs/src/api.md
```diff
@@ -139,4 +139,4 @@ The following is not part of the public API.
 Modules = [DifferentiationInterface]
 Public = false
 Filter = t -> !(Symbol(t) in [:outer, :inner])
-```
+```
```
gdalle: Does this do anything? If not, can we remove it to minimize the diffs without meaning?

DifferentiationInterfaceTest/Project.toml
```diff
@@ -50,6 +50,7 @@ FiniteDifferences = "0.12"
 Flux = "0.16"
 ForwardDiff = "0.10.36"
 Functors = "0.4, 0.5"
+GTPSA = "1.3.1"
```
gdalle:
```suggestion
```

DifferentiationInterfaceTest/Project.toml
```diff
@@ -78,6 +79,7 @@ FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+GTPSA = "b27dd330-f138-47c5-815b-40db9dd9b6e8"
```
gdalle:
```suggestion
```

DifferentiationInterface/test/Back/GTPSA/test.jl
```diff
@@ -0,0 +1,26 @@
...
+
+LOGGING = get(ENV, "CI", "false") == "false"
+
+for backend in [AutoGTPSA()]
+    @test check_available(backend)
+    @test check_inplace(backend)
+end
+
+# Test no Descriptor (use context)
+test_differentiation(AutoGTPSA(); type_stability=:full, logging=LOGGING);
```
gdalle, line 19: Congrats on making everything type-stable!!!

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/DifferentiationInterfaceGTPSAExt.jl
```diff
@@ -0,0 +1,13 @@
+module DifferentiationInterfaceGTPSAExt
+
+import DifferentiationInterface as DI
+using ADTypes: AutoGTPSA
+using GTPSA
+using LinearAlgebra
```
gdalle: Can we use explicit imports here too and add the same test as for other backends?

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -138,6 +141,15 @@ Nothing specific to mention.
...
 For higher level operators, preparation creates a [config object](https://juliadiff.org/ForwardDiff.jl/stable/user/api/#Preallocating/Configuring-Work-Buffers), which can be type-unstable.
 
+### GTPSA
+
+For all operators, preparation preallocates the input [`TPS`s](https://bmad-sim.github.io/GTPSA.jl/stable/man/d_tps/), and for in-place functions the output `TPS`s as well. For minimal allocations of `TPS` temporaries inside of a function, the [`@FastGTPSA`/`@FastGTPSA!`](https://bmad-sim.github.io/GTPSA.jl/stable/man/k_fastgtpsa/) macros are recommended.
+
+If a GTPSA [`Descriptor`](https://bmad-sim.github.io/GTPSA.jl/stable/man/c_descriptor/) is not provided to `AutoGTPSA`, then a `Descriptor` will be generated in preparation based on the context.
+
+!!! danger
+    When providing a custom GTPSA `Descriptor` to `AutoGTPSA`, it is the responsibility of the user to ensure that the number of variables + parameters is not less than the number of inputs of the provided function. Undefined behavior and crashes may occur if this is not the case.
```
gdalle: What is meant here by "variables + parameters"? Can you clarify?

gdalle: In DI parlance, is it `Constant`s?

DifferentiationInterface/test/Back/GTPSA/test.jl
```diff
@@ -0,0 +1,26 @@
...
+
```
gdalle, line 20: Should we test on the brand new `complex_scenarios` too?

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/twoarg.jl
```diff
@@ -0,0 +1,195 @@
...
+
+    return GTPSATwoArgJacobianPrep(xt, yt)
+end
+
+function DI.jacobian(
+    f!, y, prep::GTPSATwoArgJacobianPrep, ::AutoGTPSA, x, contexts::Vararg{DI.Context,C}
+) where {C}
+    foreach((t, xi) -> t[0] = xi, prep.xt, x) # Set the scalar part
+    fc! = DI.with_contexts(f!, contexts...)
+    fc!(prep.yt, prep.xt)
```
gdalle, line 148: What is the purpose of this call?

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/twoarg.jl
```diff
@@ -0,0 +1,195 @@
...
+# or a vector of pre-allocated TPSs.
+#
+# Output: Contains a vector of pre-allocated TPSs
+struct GTPSATwoArgPushforwardPrep{X,Y} <: DI.PushforwardPrep
+    xt::X
+    yt::Y
+end
+
+function DI.prepare_pushforward(
+    ::F, y, backend::AutoGTPSA{D}, x, tx::NTuple, ::Vararg{DI.Context,C}
```
gdalle: If I'm not mistaken, `DI.Cache` is not supported as a `Context` subtype, so I would replace all annotations `DI.Context` with `DI.Constant`?

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/twoarg.jl
```diff
@@ -0,0 +1,195 @@
...
+
+function DI.value_and_jacobian(
+    f!,
+    y,
+    prep::GTPSATwoArgJacobianPrep,
+    backend::AutoGTPSA,
+    x,
+    contexts::Vararg{DI.Context,C},
+) where {C}
+    jac = DI.jacobian(f!, y, prep, backend, x, contexts...)
```
gdalle: How do we ensure that `y` contains the right thing at the end of `GTPSA.jacobian!`?

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/onearg.jl
```diff
@@ -0,0 +1,634 @@
...
+        d = backend.descriptor
+    else
+        d = Descriptor(length(x), 1) # n variables to first order
+    end
+
+    # We set the slopes of each variable to 1 here, this will always be the case for gradient
+    if x isa Number
+        xt = TPS{promote_type(typeof(x), Float64)}(; use=d)
+        xt[1] = 1
+        return GTPSAOneArgGradientPrep(xt)
```
gdalle: `gradient` cannot be called on `x::Number`, hence the lack of coverage for these lines (which can be deleted)

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/onearg.jl
```diff
@@ -0,0 +1,634 @@
...
+end
+
+function DI.gradient(
+    f, prep::GTPSAOneArgGradientPrep, ::AutoGTPSA, x, contexts::Vararg{DI.Context,C}
+) where {C}
+    foreach((t, xi) -> t[0] = xi, prep.xt, x) # Set the scalar part (slopes set in prepare)
+    fc = DI.with_contexts(f, contexts...)
+    yt = fc(prep.xt)
+    if prep.xt isa Number
+        return yt[1]
```
gdalle: Same remark on `Number`

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/onearg.jl
```diff
@@ -0,0 +1,634 @@
...
+function DI.hvp(
+    f,
+    prep::GTPSAOneArgHVPPrep,
+    backend::AutoGTPSA,
+    x,
+    tx::NTuple,
+    contexts::Vararg{DI.Context,C},
+) where {C}
+    DI.hessian!(f, prep.hess, prep.hessprep, backend, x, contexts...)
+    #return prep.hess
```
gdalle:
```suggestion
```

DifferentiationInterface/ext/DifferentiationInterfaceGTPSAExt/onearg.jl
```diff
@@ -0,0 +1,634 @@
...
+    GTPSA.gradient!(grad, yt; include_params=true, unsafe_inbounds=true)
+    unsafe_fast = D == Nothing ? true : false
+    GTPSA.hessian!(
+        prep.hess,
+        yt;
+        include_params=true,
+        unsafe_inbounds=true,
+        unsafe_fast=unsafe_fast,
+        tmp_mono=prep.hessprep.m,
+    )
```
gdalle: Maybe we can use `DI.value_gradient_and_hessian!`?

gdalle (changes requested):
Thank you and sorry for the delay! Here are my remarks.
I don't understand 100% of the code but I'm confident enough in my test suite and in your work, so if the last few comments are addressed, I will happily merge

gdalle:
Note that you can commit the changes in a batch instead of one by one, by going to the "Files changed" tab

mattsignorelli:
OK thank you for your thorough review!

The only remaining points to check are:
1. Documentation of "variables" is sufficient
2. That `gradient` can take `Any`?
3. Clarification on the `Context` question

gdalle:
> Documentation of "variables" is sufficient

Yeah, I think the link to the GTPSA docs is enough. Btw, I was wondering what the point of `include_params=true` is?

> That gradient can take Any?

For reverse-mode backends, yes, `gradient` applies to an arbitrary struct. For forward-mode backends, we can only guarantee its behavior on `AbstractArray` (that's what ForwardDiff offers). So typing `x::AbstractArray` is fine.

> Clarification on the Context question

Yeah I think you should replace `Context` with `Constant` everywhere

mattsignorelli:
> Yeah, I think the link to the GTPSA docs is enough. Btw, I was wondering what the point of `include_params=true` is?

GTPSA also allows one to specify differentials for "parameters", and those can be truncated at a different order than for the "variables". The `include_params` kwarg just ensures that the result includes the partial derivatives for those differentials. 

The distinguishing of "variables" and "parameters" is because if you have say a 3rd order map representing the transport of some particle beam through a machine ("variables" are deviations from the reference particle orbit), and then you can optimize certain higher order partial derivatives of this map using "parameters" to just first order. 

> For reverse-mode backends, yes, `gradient` applies to an arbitrary struct. For forward-mode backends, we can only guarantee its behavior on `AbstractArray` (that's what ForwardDiff offers). So typing `x::AbstractArray` is fine.

Ah, I understand now. I'll remove these lines for 100% coverage then

> Yeah I think you should replace `Context` with `Constant` everywhere

OK!

mattsignorelli:
This should be ready to go

gdalle (approved):
thank you and congrats on seeing this through!

PR #341, opened 2024-06-27
Duplicated f for Enzyme to handle closures

### LINKED ISSUE #339: Make Enzyme dispatches compatible with closures

gdalle:
Fix #339


wsmoses:
This has highly non-trivial implications -- especially if you only support multi arguments etc via closures.

This will likely result in significant perf downgrades for folks, please do not merge if you want to maintain  performance for downstream users.

gdalle:
Superseded by #375

Issue #311, opened 2024-06-06
Multiple arguments / activities?


gdalle:
At the moment the only viable way I see is to support one active and one inactive argument, and leave it to ComponentArrays to combine several arguments if necessary. It could be done by passing a tuple (x, c) instead of the input x. @wsmoses do you think that might get us 90% of the way to where we need to be?

See the following discussions:

- #307
- https://github.com/LuxDL/Lux.jl/issues/685
- https://github.com/SciML/OptimizationBase.jl/pull/54
- https://github.com/tpapp/LogDensityProblemsAD.jl/pull/29

wsmoses:
That would be an improvement but we need to make sure that the decomposition of the arguments doesn't result in a materialized intermediate variable

gdalle:
Can you give an example and explain why that's a bad thing?

wsmoses:
On movile so link chasing is hard but the logprovlemsAD has a thread where
I described why closures were bad. The intermediate variable — whether
tuple struct or closure — will have the same impact if not expanded

gdalle:
Yeah I added that link above, I guess I'm just wondering what you mean by intermediate variable in this setting. I don't see why grouping multiple arguments into one or unpacking them would require a closure

jbrea:
> and leave it to ComponentArrays to combine several arguments if necessary

Not sure, this would work for my use case. Typically, I have a function `logp(data, model, parameters)`. `data` is constant and usually a vector of named tuples of whatever. `model` is usually a structure that contains configuration and state variables. The state variables need to be shadowed, as they change, but I don't care about their derivatives. The only thing I care about are the partial derivatives of `logp` with respect to `parameters`. Maybe it would make sense to follow `Lux` style and separate the model into two parts: constant configuration and non-contant state, which would result in `logp(data, config, state, parameters)`. Would it really make sense to put `data, model` or `data, config, state` into a `ComponentArray`? How much would it limit the way `data`, `config` and `state` can be written?

To be clear: I don't care, if DifferentiationInterface covers my use case or not. I'm fine with using Enzyme directly. I was just exploring this direction and thought I report back my experience :slightly_smiling_face: .

gdalle:
Thanks for your feedback! I imagine the following people may also have opinions:

- @avikpal for Lux
- @oxinabox for Zygote/Diffractor
- @willtebbutt for Tapir
- @adrhill forever <3

gdalle:
[...]

PR #344, opened 2024-07-04
Add Enzyme performance warning

gdalle:
**DI README**

- Add warning about Enzyme performance pitfalls, until we solve #311 and #339. @wsmoses what do you think?

### DIFF
--- a/DifferentiationInterface/README.md
+++ b/DifferentiationInterface/README.md
@@ -48,6 +48,10 @@ We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/
 - [Tracker.jl](https://github.com/FluxML/Tracker.jl)
 - [Zygote.jl](https://github.com/FluxML/Zygote.jl)
 
+Note that in some cases, using DifferentiationInterface.jl will be slower than calling backend-specific utilities.
+This is mostly true for Enzyme.jl, whose handling of activities and multiple arguments unlocks additional performance compared to our simple API.
+We are working on this challenge, and welcome any suggestions or contributions.
+
 ## Installation
 
 To install the stable version of the package, run the following code in a Julia REPL:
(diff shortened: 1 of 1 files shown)


DifferentiationInterface/README.md
```diff
@@ -48,6 +48,10 @@ We support all of the backends defined by [ADTypes.jl](https://github.com/SciML/
 - [Tracker.jl](https://github.com/FluxML/Tracker.jl)
 - [Zygote.jl](https://github.com/FluxML/Zygote.jl)
 
+Note that in some cases, using DifferentiationInterface.jl will be slower than calling backend-specific utilities.
+This is mostly true for Enzyme.jl, whose handling of activities and multiple arguments unlocks additional performance compared to our simple API.
```
wsmoses: Also add closures specifically and perhaps something along the lines of "for users who want to differentiate such code with Enzyme, consider using Enzyme.autodiff / Enzyme.gradient / Enzyme.jacobian / Enzyme.hvp directly"

gdalle: Okay for closures, but to recommend Enzyme syntactic sugar I'd love for it to pass all my DI tests. Last time I checked, there were still cases where the Enzyme version is wrong, eg https://github.com/EnzymeAD/Enzyme.jl/issues/1391.

wsmoses: Oh sorry missed that issue, but then perhaps just say autodiff / gradient / hvp directly then?

My guess for your issue looking at it is that it does have all the correct terms but per row major col major stuff it’s row major in one part col in other (as Julia is col major iirc)

gdalle: Yeah that might be it. My point is: as soon as there is an operator available in Enzyme, I try to use it in DI too instead of reinventing the wheel. Assuming this Jacobian issue gets fixed and I also integrate your new HVP operator without troubles, I'm okay with adding a pointer to the Enzyme syntax 😊 sound good?

wsmoses: oh also we should add something like "Some functions are differentiable with Enzyme directly, but not with DI". Since this one feels more critical to know, maybe we just write "consider using Enzyme directly" and update this to add gradient/jacobian/hvp/etc once those are in?

PR #347, opened 2024-07-13
Update operator doc strings

adrhill:
* Refer to preparation functions in all operators via `document_preparation` function in `utils/printing.jl`
* Document alternative names for pushforwards and pullbacks


gdalle (approved):
lgtm!

gdalle:
@adrhill can you check that it renders correctly?

Go to https://documenter.juliadocs.org/stable/lib/public/#Documenter.deploydocs and check out the URL used by `push_preview`.
In DifferentiationInterface it's slightly different, so the link is https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/previews/PR347/

PR #352, opened 2024-07-13
First test scenarios for Flux gradients

### LINKED ISSUE #346: Add neural network gradient tests [Juliacon 2024 Hackathon]
**Goal**

High-level: Support structures beyond arrays and numbers in DifferentiationInterface.

Low-level: Write tests for taking gradients of neural networks with Flux and Lux.

**Steps**

1. Read the slides of my autodiff intro https://gdalle.github.io/JuliaCon2024-AutoDiff/#/title-slide
2. Read some of the DifferentiationInterface documentation to understand the general ideas: 
    - [DI tutorial](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/tutorial1/)
    - [DI operators](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/operators/)
    - [DITest tutorial](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterfaceTest/stable/tutorial/)
    - [DITest Scenario API](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterfaceTest/stable/api/#DifferentiationInterfaceTest.Scenario)
3. Read [DITest source code](https://github.com/gdalle/DifferentiationInterface.jl/blob/60cdb043fe58ef3431513c90a19e20846d979614/DifferentiationInterfaceTest/src/scenarios/default.jl#L160-L246) to understand how test scenarios are defined
4. Fork the DifferentiationInterface repository and open a pull request
5. Add a file to [`DifferentiationInterfaceTest/src/scenarios`](https://github.com/gdalle/DifferentiationInterface.jl/tree/main/DifferentiationInterfaceTest/src/scenarios) called `flux.jl`
6. Define a `GradientScenario` involving a very simple neural network built with [Flux.jl](https://github.com/FluxML/Flux.jl), for instance the one in [this tutorial](https://fluxml.ai/Flux.jl/stable/guide/models/basics/#Stacking-It-Up).
7. Usually in deep learning we differentiate with respect to the parameters of a layer. In Flux, these parameters are stored within the layer itself. So the gradient we need is the gradient of `layer(input)` with respect to `layer`!!! In other words, for your `GradientScenario`, you will have `f(layer) = layer(fixed_input)` as the function (it only applies the layer to a fixed input).
8. Add a file to [`DifferentiationInterface/test/Single/Zygote`](https://github.com/gdalle/DifferentiationInterface.jl/tree/main/DifferentiationInterface/test/Single/Zygote) called `flux.jl` and test your scenario with [`DifferentiationInterfaceTest.test_differentiation`](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterfaceTest/stable/api/#DifferentiationInterfaceTest.test_differentiation). Take inspiration from the other test files.
9. Since `layer` is not an array, the returned type will not be an array either: the gradient will be some form of Flux layer as well (I think), so you probably want to compute the ground truth with Zygote at first to see how it is structured.

**If you need help**

- Come find me (@gdalle)
- Check out https://modernjuliaworkflows.github.io/ for the Julia development tricks
- Check out https://kshyatt.github.io/post/firstjuliapr/ for the GitHub-related issues

**Participants**

- @nialamarcotte

nialamarcotte:
Work done during JuliaCon Hackathon; preparing test for neural networks


gdalle:
Thanks so much for your contribution! It may take a few days but I'll get back to you, give you some feedback on the code and take it from here if that's okay. I'm sorry that the task was so daunting, but very grateful that you agreed to give it a try :) And if you need any more resources on AD, let me know

gdalle:
**CI**

- [x] Adjust groups



**DIT source**

- [x] Support custom `isequal` function in correctness tests, in addition to `isapprox`
- [x] Adjust `mycopy_random` and `mysimilar` for arbitrary structures with Functors

**DIT extensions**

- [x] Add Flux extension with gradient scenarios, whose correctness is checked against FiniteDifferences (see https://github.com/FluxML/Flux.jl/issues/2469#issuecomment-2228696472)
- [x] Add more scenarios with various types of networks

**DIT tests**

- [x] Test new Flux scenarios with Zygote

**DI extensions**

- [x] In the Enzyme extension, support non-array inputs in reverse mode:
  - Relax type constraints
  - Avoid querying `eltype(x)`, backpropagate `dy=true` instead

**DI tests**

- [x] Test new Flux scenarios with Zygote (Enzyme fails)

gdalle:
@CarloLucibello I got the first batch of tests working with Zygote. The syntax is a bit counter-intuitive because DI doesn't yet support multiple arguments, so I have to put the data `x` inside the loss closure, but it gets the job done for now.
The Enzyme tests [error](https://github.com/gdalle/DifferentiationInterface.jl/actions/runs/9961343185/job/27522654271?pr=352) on two cases though, the ones involving `Flux.Scale` and `Flux.SkipConnection`. Any idea what could be different about those two? @wsmoses could this be one of the cases where we want to `Duplicate` the function `f` itself for the time being?
What other backends do we want to test here? Tracker?

wsmoses:
Actually this is probably a case of exactly the opposite in which doing so would be harmful. Marking the closure as differentiable would mean that Enzyme would also end up computing the derivative of the input [and then throwing it away].

wsmoses:
Per the error message, it looks like the closure is causing issues where it can't be compiled due to mixed activity [e.g. one element of the closure being differentiable the other not]. As the error message states, you can enable runtime activity to resolve at a slight [but not asymptotic performance loss]. But also using multiple arguments here in Enzyme directly would also presumably resolve this

gdalle:
> Per the error message, it looks like the closure is causing issues where it can't be compiled due to mixed activity [e.g. one element of the closure being differentiable the other not]. As the error message states, you can enable runtime activity to resolve at a slight [but not asymptotic performance loss].

I tried enabling runtime activity but the stack traces just got angrier 😱 

> But also using multiple arguments here in Enzyme directly would also presumably resolve this

Agreed, but setting this up in the single-argument case also helped me resolve some other issues to pave the way for non-array testing, so it's a great first step. For the multi-argument case, what I need most of all is inspiration and ideas to improve my initial API proposal at https://github.com/gdalle/DifferentiationInterface.jl/issues/311#issuecomment-2228066950. The idea is to be slightly different than Enzyme, and closer to the mathematical concepts that users may be familiar with.

gdalle:
Merging this without the Enzyme tests because there's already some useful stuff in there. Billy if you have a clue what the problem was in https://github.com/gdalle/DifferentiationInterface.jl/actions/runs/9963932557/job/27531092106 with runtime activity switched on, I'd love to hear your thoughts

PR #360, opened 2024-07-16
Use Zygote Buffer for mutating functions

### LINKED ISSUE #348: Support `f!(y, x)` for Zygote with `Zygote.Buffer`
See https://fluxml.ai/Zygote.jl/stable/utils/#Zygote.Buffer

Ping @Vaibhavdixit02

gdalle:
**DI extensions**

- Use `Zygote.Buffer` to allow writing into the output `y` when differentiating through `f!(y, x)` with Zygote. Note that other mutations inside the function (e.g. of temporary storage) will still error.


gdalle:
@Vaibhavdixit02 I can't figure out why this fails and it is very low-priority for me. If you wanna get your hands dirty, feel free, but otherwise I don't know if the issue is really worth pressing?

Vaibhavdixit02:
I have figured out a way to do it at my end so I don't need this as much anymore, I'll give this a shot but if you couldn't get it to work I don't believe that I'll be able to in your codebase haha

Issue #343, opened 2024-07-01
Tests on generic structs


gdalle:
What kind of structs should we add to enable deep learning applications?

adrhill:
The answer will be different for Flux and Lux, the former most likely requiring support for [Functor.jl](https://github.com/FluxML/Functors.jl).
It's worth noting that parameters in DL are somewhat of a mess: some are trainable arrays, some are trainable scalar values and other are non-trainable parameters, e.g. [moving statistics in `BatchNorm` layers](https://github.com/FluxML/Flux.jl/blob/39d04ebea7fe8fcc2eabb9656135cef674f0fe88/src/layers/normalise.jl#L318-L319).

avik-pal:
> It's worth noting that parameters in DL are somewhat of a mess: some are trainable arrays, some are trainable scalar values and other are non-trainable parameters, e.g. [moving statistics in BatchNorm layers](https://github.com/FluxML/Flux.jl/blob/39d04ebea7fe8fcc2eabb9656135cef674f0fe88/src/layers/normalise.jl#L318-L319).

This is what Lux and its derivative frameworks were designed to fix. `LuxCore.jl` is basically the interface specification that Lux models need to abide by. Lux simply specifies bindings to `LuxLib`/`NNlib` and some utilities users need for deep learning abiding by those specifications. To summarize the interface specification, you have 4 parts:

1. `model`: This is an immutable structure of the problem (more specifically, neural network architecture). **Values Never Participate in Gradient Calculation**
2. `ps`: Parameters. All of these are trainable. They may contain scalars but generally don't. When we want to compute the gradients, this is what we are typically computing the gradients for.
3. `st`: States (can think of them as non-trainable parameters). These can be anything like arrays, Vals, Numbers, etc. **Values Never Participate in Gradient Calculation**
4. `data`: You might need to compute the gradients for this if users want it. But typically, this is not required.

> The answer will be different for Flux and Lux, the former most likely requiring support for [Functor.jl](https://github.com/FluxML/Functors.jl)

Lux uses a restrictive definition of `fmap` (for operations on parameters) for type-stability (https://lux.csail.mit.edu/stable/api/Lux/utilities#Lux.recursive_map) but `fmap` is the general solution for both.

willtebbutt:
[...]

Issue #105, opened 2024-03-26
Testing NNlib / Lux / Flux


gdalle:
Lower hanging fruit: NNLib.jl, because there are less weird structs, mostly arrays

Cross-referencing:

- #346 
- https://github.com/LuxDL/Lux.jl/issues/769
- https://github.com/FluxML/Flux.jl/issues/2469

gdalle:
**Slow**
- Replace several calls to `grad_test` with a vector of scenarios, like in `softmax.jl` and then `scatter.jl`

**Fast**
- Replace `grad_test` and watch the world crumble: https://github.com/FluxML/NNlib.jl/blob/master/test/test_utils.jl

avik-pal:
If we want to be adventurous, you can change https://github.com/LuxDL/LuxTestUtils.jl and all downstream CPU tests in Lux will be triggered (and we just need to copy one of the buildkite files from LuxLib to trigger the CUDA + AMDGPU tests)

gdalle:
Don't tempt me Avik

avik-pal:
On a serious note though, I had to write it to mostly deal with arrays or at least convert structures to arrays https://github.com/LuxDL/LuxTestUtils.jl/blob/143a51f0d2fb4cbc75ea583c706ff5194be103d2/src/LuxTestUtils.jl#L387-L398, so that could be helpful to writing your test suite. (But this is also terribly inefficient and only tests correctness and definitely don't combine `@test_gradients` with `@jet`)

gdalle:
Are the tests of LuxTestUtils already interesting to run locally, or should we wait for the Downstream CI every time?

avik-pal:
no the tests there do nothing practically, it is all via the downstream CI

avik-pal:
but the Lux test suite doesn't take long -- 10 mins on a nicer machine (like the buildkite ones) but github actions ones take longer ~30 mins

If you want to test locally, set `RETESTITEMS_NWORKERS` and it will be much faster

gdalle:
So the workflow is to:
1. fork LuxTestUtils.jl and Lux.jl
2. put my own gradient callers in LuxTestUtils.jl
3. dev LuxTestUtils.jl into the test environment of Lux.jl
4. test Lux.jl

right?

avik-pal:
If you want to test locally yes.

gdalle:
[...]

PR #372, opened 2024-07-19
Add Lux tests

gdalle:
**DIT extensions**

- [x] Set up a Lux extension to define scenarios, modeled after the [Flux one](https://github.com/gdalle/DifferentiationInterface.jl/pull/352)
- [x] Define the scenarios (out-of-place gradient with respect to parameters), their ground truth values (FiniteDifferences) and comparison functions (`LuxTestUtils.check_approx`)

**DIT tests**

- [x] Test Lux scenarios with one backend (Zygote)

**DI tests**

- [x] Set up a folder `test/Down/Lux` for Lux tests and add it to CI
- [x] Test Lux scenarios with one backend (Zygote)

---

See also:

- https://github.com/LuxDL/Lux.jl/issues/769
- https://github.com/gdalle/DifferentiationInterface.jl/issues/105
- https://github.com/gdalle/DifferentiationInterface.jl/issues/343


avik-pal:
> Gradient with respect to parameter only? Or parameter + state in a ComponentVector?

The latter won't work because state can contain non-array-like objects like `Val`.

> Is Zygote the ground truth or Finite Differences?

Its Finite Differences (except in edge cases like Float16 testing where the error is quite high for finite differencing). I left more detailed comment on the issue on Lux

PR #375, opened 2024-07-22
Handle closures with Enzyme

### LINKED ISSUE #339: Make Enzyme dispatches compatible with closures

gdalle:
**DI extensions**

- Handle the new `AutoEnzyme(constant_function=false)` by providing `Duplicated(f, df)` instead of `Const(f)` to `autodiff`
- Restrict calls to high-level API to `constant_function=true` because of https://github.com/EnzymeAD/Enzyme.jl/issues/1670
- At the moment, `constant_function=false` will fail for functions that don't close over mutable data. It would require `Active` or `MixedDuplicated` to make it work, see https://github.com/EnzymeAD/Enzyme.jl/issues/1669

**DIT source**

- Implement `make_closure(scenario)` which creates a closure over `x` and `y` buffers (differentiable data) and calls it instead

**DIT tests**

- Test the closure scenarios with FiniteDiff (they would fail with ForwardDiff for lack of a Dual cache)

**DI tests**

- Add tests over closure scenarios for `AutoEnzyme(constant_function=false)`

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -44,7 +44,7 @@ DifferentiationInterfaceTrackerExt = "Tracker"
 DifferentiationInterfaceZygoteExt = ["Zygote", "ForwardDiff"]
 
 [compat]
-ADTypes = "1.5.0"
+ADTypes = "1.6.1"
 ChainRulesCore = "1.23.0"
 Compat = "3,4"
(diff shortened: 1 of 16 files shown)


gdalle:
Regardless of which setting ends up being the default in ADTypes, I could use your help on this one @wsmoses

PR #383, opened 2024-07-25
[WIP] Setup tape for pullback for ForwardDiff over ReverseDiff

### LINKED ISSUE #321: Tape-able pullback with ReverseDiff
At the moment, the ReverseDiff pullback is a bit hacky because it takes the gradient of a closure. Ideally we should hook into ReverseDiff internals

Vaibhavdixit02:
Initial draft to sketch out what I think it should look like. Fixes #321

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceReverseDiffExt/DifferentiationInterfaceReverseDiffExt.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceReverseDiffExt/DifferentiationInterfaceReverseDiffExt.jl
@@ -23,7 +23,8 @@ using ReverseDiff:
     hessian,
     hessian!,
     jacobian,
-    jacobian!
+    jacobian!,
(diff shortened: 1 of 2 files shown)


gdalle:
Wait I don't understand, pullback is a first order operator, what does it have to do with ForwardDiff over ReverseDiff?
What we need are two things:

- an HVP with ForwardDiff over ReverseDiff
- a pullback with ReverseDiff that uses the existing tape mechanism

Which are you trying to tackle?

Vaibhavdixit02:
>a pullback with ReverseDiff that uses the existing tape mechanism

As I understand the current pullback needs to compile the tape everytime and cannot use the preparation in the presence of the Duals from ForwardDiff, so this let's you avoid that by adding a preparation and pullback that can work with Duals.

The hvp gets called fine even now other than the recompilation so I am not sure what you mean needs to be done there.

PR #402, opened 2024-08-05
Put local installation in CI

gdalle:
**CI**

- Add CI script to install local version of packages instead of doing it in the `runtests.jl` (inspired by GraphNeuralNetworks.jl, see https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/433)

**Tests**

- Put `LOGGING` variable in every file to make them independent


gdalle:
https://discourse.julialang.org/t/how-to-use-local-version-of-test-dependency/117833

PR #404, opened 2024-08-07
Set up test env fully in CI script

gdalle:
**DI and DIT test**

- Move test deps from main `Project.toml` to `test/Project.toml` so that the `test/Manifest.toml` can contain the correct path

**CI**

- Update test CI to dev DIT into `DI/test`


gdalle:
And we get the infamous https://github.com/JuliaLang/Pkg.jl/issues/1585.
Note however that CI works on 1.11, probably because https://github.com/JuliaLang/Pkg.jl/pull/3276 was merged in the meantime.

Issue #413, opened 2024-08-17
StaticArrays woes


ExpandingMan:
There currently seem to be quite a few issues with StaticArrays on many different back-ends.  In most cases this is inefficiency due to inappropriate allocations (sometimes quite severe), in other cases there are outright errors.  This issue is to document the various problems.  Note that very few, if any of these are actually issues with DifferentiationInterface.jl itself, but rather with the back-ends.

In what follows we will use
```julia
f1(x) = sum(x .* x)
f2(x) = x .* x

testarray() = @SVector ones(4)

x = testarray()
```

### Improper Allocation in Gradient
```
◖◗ @btime gradient($f1, $(AutoEnzyme()), $x)
  107.511 ns (9 allocations: 688 bytes)
4-element SVector{4, Float64} with indices SOneTo(4):
 2.0
 2.0
 2.0
 2.0
```
This is likely due to insufficient specialization in `Enzyme.gradient` for StaticArrays.  I have confirmed that a raw `Enzyme.autodiff` is efficient and does not allocate.  I'm attempting to address this, among other things [in this PR](https://github.com/EnzymeAD/Enzyme.jl/pull/1733).

### Invalid Construction in Jacobian
```
◖◗ @btime jacobian($f2, $(AutoEnzyme()), $x)
ERROR: DimensionMismatch: No precise constructor for SVector{4, Float64} found. Length of input was 1.
Stacktrace:
  [1] _no_precise_size(SA::Type, x::Tuple{Float64})
    @ StaticArrays ~/.julia/packages/StaticArrays/MSJcA/src/convert.jl:169
  [2] length_match_size(::Type{SVector{4, Float64}}, x::Tuple{Float64})
    @ StaticArrays ~/.julia/packages/StaticArrays/MSJcA/src/convert.jl:138
  [3] adapt_size(::Type{SVector{4, Float64}}, x::Tuple{Float64})
    @ StaticArrays ~/.julia/packages/StaticArrays/MSJcA/src/convert.jl:107
  [4] construct_type(::Type{SVector{4, Float64}}, x::Tuple{Float64})
    @ StaticArrays ~/.julia/packages/StaticArrays/MSJcA/src/convert.jl:95
  [5] Type
    @ ~/.julia/packages/StaticArrays/MSJcA/src/convert.jl:175 [inlined]
  [6] #3
[...]

PR #414, opened 2024-08-18
Unprepared operators for ForwardDiff

gdalle:
**Versions**

- Bump DI to v0.5.14

**Extensions**

- ForwardDiff: Add direct implementations for unprepared operators that do not call preparation first (see #413).

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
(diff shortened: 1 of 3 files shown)


ExpandingMan:
For posterity note that it should probably be considered a bug in ForwardDiff that these are needed, since most of the inefficiency in the prepared methods seems to be coming from type instability on the part of ForwardDiff, but considering that a fix for that is not likely to come very soon this is certainly nice to have.

gdalle:
Apparently, ForwardDiff's [StaticArrays extension](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/ext/ForwardDiffStaticArraysExt.jl) just ignores the config object which I create during preparation, and whose preparation causes type instability. With these new unprepared functions we should hit the right shortcuts

PR #464, opened 2024-09-12
Update Tapir to use cached rules where possible

willtebbutt:
We've recently added a feature to Tapir.jl which does some (very conservative) rule caching. In order to take advantage of this, I've upgraded the Tapir extension.

(The plan is still to rename Tapir.jl to something else, it's just taking a while. In the meantime, I'd like to ensure that this feature is rolled out).


gdalle:
Cool! The tests are probably gonna fail because Enzyme is acting up (https://github.com/EnzymeAD/Enzyme.jl/issues/1804), once this is solved we'll re-run.
Note that v0.6 still needs a bit of work so this won't get tagged right away. It's a breaking change and I wanna make sure I get it right.

willtebbutt:
Sounds good to me -- there's no particular rush because (I believe) this is strict improvement over what we already have.

Would you like me to remove the patch bump and just incorporate this into the 0.6 release?

gdalle:
Yeah we might as well.

gdalle (changes requested):
This looks good, can you just bump the weakdep compat to the earliest version of Tapir supporting this?

gdalle (approved):

Issue #312, opened 2024-06-06
Vector / chunked / batch mode


gdalle:
This one is rather easy in principle: just wrap the JVP/VJP/HVP seed in a Batch struct that is essentially a named tuple, and dispatch on that

gdalle:
Partially solved by #320, still needs:

- [x] Batched HVPs based on batched pushforward/pullback (done in #330 except for reverse-over-forward)
- [x] Batched pushforward with ForwardDiff (done in #328)
- [ ] Batched pushforward and pullback with Enzyme
- [ ] Tests on efficiency (type stability, allocations)
- [x] Independent tests of the primitive operators with batched scenarios (done in #331)

gdalle:
It would be nice to avoid the distinction between `operator` and `operator_batched`, but that would require wrapping inputs into a `Single` struct or so

gdalle:
Further improved by #412, which makes batch mode the default that extensions need to implement

gdalle:
Batch mode for Enzyme is nearly complete since #471, it only needs https://github.com/EnzymeAD/Enzyme.jl/issues/1883 to work with out-of-place functions

PR #471, opened 2024-09-18
Adapt to Enzyme v0.13

### LINKED ISSUE #468: Adapt to pending Enzyme breaking change
EnzymeRules for forward mode are defined differently [not relevant here]
ForwardMode moved DupNoNeed / Dup to specify if the return is needed or not -> Forward and ForwardWithPrimal

Runtime Activity is now in the mode

gradient/jacobian all optionall return values with the WithPrimal arg

ExpandingMan:
EDIT by @gdalle: see [final description below](https://github.com/gdalle/DifferentiationInterface.jl/pull/471#issuecomment-2367996453)

I'm a little bewildered by testing of this package, so I haven't checked this thoroughly yet and likely something somewhere will break, but it seems to mostly work.


gdalle:
Yeah the testing is a bit chaotic because I have a separate GitHub Actions workflow and a separate set of dependencies for each AD backend. Otherwise it's very hard to instantiate an environment with all 12 of them, especially since some are nearly unmaintained. As a result, testing locally kinda sucks, but if you become more involved in DI development we may be able to figure out something smarter.
Thanks for this contribution in any case! I'll try to take it from here, but do you happen to have a summary of the breaking changes in Enzyme v0.13? I didn't see one on the repo's release notes.

ExpandingMan:
I have not seen an explicit list of breaking changes, I just went through and tested all the basic cases myself.  Seems to be almost done, but as you can see it's not quite there.

Note that, as per our conversation on slack, this is going to have to flatten the Enzyme jacobians, which, as of writing, is not in yet.

ExpandingMan:
I now have this flattening the jacobians.  Tests now seem to mostly pass, but there are still some cases where it's mysteriously returning all zeros.

Btw, it's unfortunate that throwing test errors doesn't actually tell you what failed, it just shows you what the result is (because it's just showing the standard `@test` failure message) and it's not particularly easy to figure out what's going on by looking in the source code.  (Just something to think about as I saw you had another PR for cleaning up testing.)

gdalle:
**DI extensions (Enzyme)**

- Remove the deferred mechanism (no longer needed)
- Implement utilities to modify the `Mode` object (with/without primal, splitting, error analysis)
- Revamp `pushforward` and `pullback` with these new modes. Use `autodiff_thunk` everywhere for out-of-place functions in reverse mode (see also https://github.com/EnzymeAD/Enzyme.jl/issues/1853).
- Implement batched versions (#312)
  - for forward mode (in-place and out-of-place functions)
  - for reverse mode (only in-place functions until https://github.com/EnzymeAD/Enzyme.jl/issues/1883)
- Adjust to new syntax for `Enzyme.gradient` and `Enzyme.jacobian`. Reshape the Jacobian accordingly when it is a higher-dimensional tensor.

**DI source**

- Remove mention of Enzyme deferred in `DI.nested` docstring (`nested` is useless for now)

**DI tests**

- Enzyme: Deactivate type stability tests (broken by Enzyme update).
- Temporarily deactivate Lux tests (until #484, because of compat conflicts)

**DIT tests**

- Temporarily deactivate Lux tests

ExpandingMan:
Thanks for finishing this.

PR #487, opened 2024-09-23
Make `DifferentiateWith` compatible with ForwardDiff, clarify docs

### LINKED ISSUE #476: Better docs for `DifferentiateWith`
- Highlight that it works with compositions
- Remove the "unstable" warning

### LINKED ISSUE #477: `DifferentiateWith` for ForwardDiff
Easy to achieve by dispatching on `Dual` and `AbstractArray{<:Dual}`

gdalle:
**DIT source**

- Clarify documentation of `DifferentiateWith`, mention compositions.

**DIT extensions**

- Make `DifferentiateWith` compatible with ForwardDiff through `Dual` and `AbstractArray{<:Dual}` overloads

**DI tests**

- Test `DifferentiateWith` for Zygote and ForwardDiff on the closurified backends, using FiniteDiff as substitute

### DIFF
--- a/DifferentiationInterface/src/misc/differentiate_with.jl
+++ b/DifferentiationInterface/src/misc/differentiate_with.jl
@@ -1,18 +1,23 @@
 """
     DifferentiateWith
 
-Callable function wrapper that enforces differentiation with a specified (inner) backend.
+Function wrapper that enforces differentiation with a substitute AD backend, different from the true AD backend that the user intends to call.
 
-This works by defining new rules overriding the behavior of the outer backend that would normally be used.
+For instance, suppose `f` is not differentiable with Zygote because it involves mutation, but you know that it is differentiable with Enzyme.
+Then `g = DifferentiateWith(f, AutoEnzyme())` is differentiable with Zygote thanks to a chain rule which calls Enzyme under the hood.
+Moreover, any composition involving `g` will also be differentiable.
 
 !!! warning
-    This is an experimental functionality, whose API cannot yet be considered stable.
-    It only supports out-of-place functions, and rules are only defined for [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible outer backends.
+    `DifferentiateWith` only supports out-of-place functions `y = f(x)` without additional context arguments.
+    It only makes these functions differentiable if the true backend is either [ForwardDiff](https://github.com/JuliaDiff/ForwardDiff.jl) or anything [ChainRules](https://github.com/JuliaDiff/ChainRules.jl)-compatible.
 
 # Fields
 
-- `f`: the function in question
-- `backend::AbstractADType`: the inner backend to use for differentiation
+- `f`: the function in question, with signature `f(x)`
+- `backend::AbstractADType`: the substitute backend to use for differentiation
+
+!!! note
+    For the substitute AD backend to be called under the hood, its package needs to be loaded in addition to the package of the true AD backend.
 
 # Constructor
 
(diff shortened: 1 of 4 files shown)


DifferentiationInterface/src/misc/differentiate_with.jl
```diff
@@ -1,18 +1,23 @@
 """
     DifferentiateWith
 
-Callable function wrapper that enforces differentiation with a specified (inner) backend.
+Function wrapper that enforces differentiation with a substitute AD backend, different from the true AD backend that the user intends to call.
```
adrhill:
```suggestion
Function wrapper that enforces differentiation with a substitute AD backend, independent from AD backends specified in calls to operators.
```

DifferentiationInterface/src/misc/differentiate_with.jl
```diff
@@ -1,18 +1,23 @@
...
 
-This works by defining new rules overriding the behavior of the outer backend that would normally be used.
+For instance, suppose `f` is not differentiable with Zygote because it involves mutation, but you know that it is differentiable with Enzyme.
+Then `g = DifferentiateWith(f, AutoEnzyme())` is differentiable with Zygote thanks to a chain rule which calls Enzyme under the hood.
+Moreover, any composition involving `g` will also be differentiable.
```
adrhill: Looks like the relationship between `f` and `g` hasn't been explained yet?

DifferentiationInterface/src/misc/differentiate_with.jl
```diff
@@ -1,18 +1,23 @@
...
 
 !!! warning
-    This is an experimental functionality, whose API cannot yet be considered stable.
-    It only supports out-of-place functions, and rules are only defined for [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible outer backends.
+    `DifferentiateWith` only supports out-of-place functions `y = f(x)` without additional context arguments.
+    It only makes these functions differentiable if the true backend is either [ForwardDiff](https://github.com/JuliaDiff/ForwardDiff.jl) or anything [ChainRules](https://github.com/JuliaDiff/ChainRules.jl)-compatible.
```
adrhill:
```suggestion
    It only makes these functions differentiable if the true backend is either [ForwardDiff](https://github.com/JuliaDiff/ForwardDiff.jl) or compatible with [ChainRules](https://github.com/JuliaDiff/ChainRules.jl).
```

DifferentiationInterface/src/misc/differentiate_with.jl
```diff
@@ -1,18 +1,23 @@
...
 
 # Fields
 
-- `f`: the function in question
-- `backend::AbstractADType`: the inner backend to use for differentiation
+- `f`: the function in question, with signature `f(x)`
+- `backend::AbstractADType`: the substitute backend to use for differentiation
+
+!!! note
+    For the substitute AD backend to be called under the hood, its package needs to be loaded in addition to the package of the true AD backend.
```
adrhill, line 25:
```suggestion
    For the substitute AD backend to be called under the hood, its package needs to be loaded in addition to the package of the backend specified in the operator.
```

DifferentiationInterface/src/misc/differentiate_with.jl
gdalle: I don't mention operators because `DifferentiateWith` also works when the true backend is not used through DI, as shown in the example

adrhill: I just find the label "true backend" misleading for the backend that isn't called.

And I think we can safely assume that people using DI's `DifferentiateWith` also use DI's interface?

gdalle: > I just find the label "true backend" misleading for the backend that isn't called.

It is called for everything that is around `g`, like the contents of `h` in the example.
It handles compositions and all the rest, only the inside of `g` is substituted.

> And I think we can safely assume that people using DI's DifferentiateWith also use DI's interface?

Not necessarily, they might be writing code for other people to use and differentiate.

adrhill:
```suggestion
Function wrapper that enforces differentiation with a substitute AD backend, independent from outer AD backend calls.
```

DifferentiationInterface/src/misc/differentiate_with.jl
adrhill:
```suggestion
    For the substitute AD backend to be called under the hood, its package first needs to be loaded.
```

DifferentiationInterface/src/misc/differentiate_with.jl
gdalle: solved

DifferentiationInterface/src/misc/differentiate_with.jl
gdalle: I thought about "inner" and "outer" but I don't want this to become confusing with the inner and outer backend of second-order AD

adrhill: I would argue that that's ok, but I understand your issues with it.

PR #500, opened 2024-09-25
[BREAKING] Transition from Tapir to Mooncake

gdalle:
> [!WARNING]
> Breaking change: `AutoTapir` is no longer supported, replaced by `AutoMooncake` (see https://github.com/SciML/ADTypes.jl/pull/89)

**DI deps**

- [x] Replace Tapir with Mooncake, bump ADTypes compat 

**DI extensions, tests, CI**

- [x] Rename everything from Tapir to Mooncake
- [x] Handle new `Mooncake.Config` (ping @willtebbutt)
- [x] Implement multi-argument support
- [x] Reactivate other tests


DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
```diff
@@ -1,42 +1,47 @@
...
+function DI.prepare_pullback(f, backend::AutoMooncake, x, ty::DI.Tangents)
     y = f(x)
     rrule = build_rrule(
-        get_tapir_interpreter(),
+        get_interpreter(),
         Tuple{typeof(f),typeof(x)};
-        safety_on=backend.safe_mode,
-        silence_safety_messages=false,
+        debug_mode=false,  # TODO: modify
+        silence_debug_messages=false,
```
willtebbutt: The intention moving forwards is that both `debug_mode` and `silence_debug_messages` will be specified via `Config`, so I would recommend getting both of these fields from `backend`.

willtebbutt: If the `backend.config === nothing`, then you should just build a `Mooncake.Config()` object via the default constructor, and pull out the required fields.

gdalle:
I think multi-argument support is operational, care to take a look? All you need to know is that:

- `contexts` can contain any number of arguments (including 0), wrapped in specific types like `Constant <: Context`
- one can access the actual data with `unwrap`, e.g. `unwrap(Constant(c)) == c`

DifferentiationInterface/README.md
```diff
@@ -37,10 +37,10 @@ We support the following backends defined by [ADTypes.jl](https://github.com/Sci
 - [FiniteDiff.jl](https://github.com/JuliaDiff/FiniteDiff.jl)
 - [FiniteDifferences.jl](https://github.com/JuliaDiff/FiniteDifferences.jl)
 - [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl)
+- [Mooncake.jl](https://github.com/withbayes/Mooncake.jl)
```
willtebbutt:
```suggestion
- [Mooncake.jl](https://github.com/compintell/Mooncake.jl)
```
The org name changed

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -144,9 +144,9 @@ For all operators, preparation generates an [executable function](https://docs.s
 !!! warning
     Preparation can be very slow for symbolic AD.
 
-### Tapir
+### Mooncake
 
-For `pullback`, preparation [builds the reverse rule](https://github.com/withbayes/Tapir.jl?tab=readme-ov-file#how-it-works) of the function.
+For `pullback`, preparation [builds the reverse rule](https://github.com/withbayes/Mooncake.jl?tab=readme-ov-file#how-it-works) of the function.
```
willtebbutt:
```suggestion
For `pullback`, preparation [builds the reverse rule](https://github.com/compintell/Mooncake.jl?tab=readme-ov-file#how-it-works) of the function.
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/DifferentiationInterfaceMooncakeExt.jl
```diff
@@ -0,0 +1,36 @@
...
+    tangent_type,
+    value_and_pullback!!,
+    zero_codual,
+    zero_fcodual,
+    zero_tangent,
+    __value_and_pullback!!
+
+function DI.check_available(::AutoMooncake)
+    return true
+end
```
willtebbutt:
```suggestion
DI.check_available(::AutoMooncake) = true
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
```diff
@@ -0,0 +1,100 @@
+struct MooncakeOneArgPullbackPrep{Y,R} <: PullbackPrep
+    y_prototype::Y
+    rrule::R
+end
+
+function DI.prepare_pullback(
+    f, backend::AutoMooncake, x, ty::DI.Tangents, contexts::Vararg{Context,C}
+) where {C}
+    y = f(x, map(unwrap, contexts)...)
```
willtebbutt, line 9: Is it part of the interface of DI that `f` must not mutate either `x` or `map(unwrap, contexts)...)`?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
```diff
@@ -0,0 +1,100 @@
...
+    f,
+    prep::MooncakeOneArgPullbackPrep{Y},
+    ::AutoMooncake,
+    x,
+    ty::DI.Tangents{1},
+    contexts::Vararg{Context,C},
+) where {Y,C}
+    dy = only(ty)
+    dy_righttype = convert(tangent_type(Y), dy)
+    new_y, (_, new_dx) = value_and_pullback!!(
```
willtebbutt, line 32: Will this work if there's a context argument? i.e. the second element of `Mooncake.value_and_pullback!!` is always a `Tuple` with as many elements as arguments to `f` + 1 (the +1 is because `f` itself gets a tangent). This looks to me like it only expects there to ever be two elements in the `Tuple`, but I'm assuming that there can be more in general.

willtebbutt (commented):
Largely looks good, just a couple of queries.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
gdalle: For now yes, because the only available context type is `Constant`. Later I'll add a second type called `Cache` with different properties.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
gdalle: I was a bit surprised but this seems to work:
```julia
julia> (a, b) = (1, 2, 3)
(1, 2, 3)

julia> b
2
```

willtebbutt: Ooooo I learned something new today!

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
willtebbutt: Cool. I ask because, of course, if you run `f` and it mutates `x`, you won't necessarily be able to call this function again (as `x` may have changed). Happy for this to be marked as resolved if this isn't considered a problem.

willtebbutt (approved):
Looks good from my perspective. I don't see how we could need to do anything else in this PR that is breaking, so if there is anything that's not been implemented optimally, we can revisit the details later on.

PR #502, opened 2024-09-25
Demonstrate Mooncake segfault

gdalle:
> [!CAUTION]
> Do not merge.


willtebbutt:
Okay, so I misunderstood what was going on here. You actually _do_ need to just use `zero_fcodual` in the two argument version, because we're not calling out to `Mooncake.value_and_gradient!!` or `Mooncake.__value_and_gradient!!` here, rather we're directly calling the rule that we generate. The rule requires that you pass in fdata, or it will segfault.

I believe that calling `zero_codual` is the correct thing to do in the single argument setting though, because you're calling `Mooncake.__value_and_pullback!!` there.

So, in short, the current code on main should be correct -- do the tests pass if you use `zero_codual`, or do you see problems?

Concerningly, `debug_mode=true` doesn't seem to catch this, so I'm going to have to dig in to why that's happening, because it's meant to catch exactly these kinds of problems, and I would have thought that this would be an easy one to catch (the fdata for the context should be `NoFData`, but it's `0.0`, which is causing the segfault).

gdalle:
I don't fully understand why, but the tests pass on main so it seems all is good for now. I'm gonna leave this PR open so that the branch doesn't disappear, just ping me when you're done investigating and I'll close it.

willtebbutt:
It's really weird -- I think maybe the problem is that the segfault occurs during inference / codegen, rather than at runtime. For context, if you give a rule which has been constructed in debug mode a `CoDual{Float64, Float64}`, it will call `Mooncake.verify_fwds_inputs`, which will in turn yield an error of the form:
```julia
ERROR: error in inputs to rule with input types Tuple{CoDual{DifferentiationInterfaceTest.MultiplyByConstant{:in, typeof(DifferentiationInterfaceTest.mat_to_mat!)}, Mooncake.NoFData}, CoDual{Matrix{Float64}, Matrix{Float64}}, CoDual{Matrix{Float64}, Matrix{Float64}}, CoDual{Float64, Float64}}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] verify_fwds_inputs(x::Tuple)
   @ Mooncake ~/.julia/packages/Mooncake/GSGyW/src/debug_mode.jl:101
 [3] top-level scope
   @ REPL[49]:1

caused by: Mooncake.InvalidFDataException("Type Float64 has fdata type Mooncake.NoFData, but got Float64.")
Stacktrace:
 [1] verify_fdata_type(P::Type, F::Type)
   @ Mooncake ~/.julia/packages/Mooncake/GSGyW/src/fwds_rvs_data.jl:285
 [2] verify_fdata_value(p::Float64, f::Float64)
   @ Mooncake ~/.julia/packages/Mooncake/GSGyW/src/fwds_rvs_data.jl:301
 [3] verify_fwds(x::CoDual{Float64, Float64})
   @ Mooncake ~/.julia/packages/Mooncake/GSGyW/src/debug_mode.jl:113
 [4] verify_fwds_inputs(x::Tuple)
   @ Mooncake ~/.julia/packages/Mooncake/GSGyW/src/debug_mode.jl:98
 [5] top-level scope
   @ REPL[49]:1
```
I confirmed that this error does indeed get thrown if you actually run `Mooncake.verify_fwds_inputs` on the inputs to `rrule`, so it must be an inference problem. Fortunately, this is quite a simple example, so I should be able to minimise it and try to figure out how to ensure that we get a sensible error if this occurs in the future.

In the mean time, if this isn't causing you problems on main, then I suggest closing this PR.

PR #513, opened 2024-09-30
Improve Mooncake caching

gdalle:
**DI extensions**

- Mooncake: use preparation to allocate tangent objects with the correct tangent type

@willtebbutt care to take a look?


DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
```diff
@@ -27,43 +39,36 @@ function DI.value_and_pullback(
     contexts::Vararg{Context,C},
 ) where {C}
     dy = only(ty)
-    dy_righttype = convert(tangent_type(typeof(y)), copy(dy))
+
+    # Set all tangent storage to zero.
+    df! = set_to_zero!!(prep.df!)
+    # dx_righttype = set_to_zero!!(prep.dx_righttype)  # TODO: why doesn't this work?
```
gdalle: For some reason when I use this line instead of the following line, the tests fail. They only fail for batch mode with 2 tangents, so my guess is that something is not being properly zero-ed/reset. Any clue?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
```diff
@@ -45,8 +49,8 @@ function DI.value_and_pullback!(
     contexts::Vararg{Context,C},
 ) where {Y,C}
     dx, dy = only(tx), only(ty)
-    dy_righttype = convert(tangent_type(Y), dy)
-    dx_righttype = set_to_zero!!(convert(tangent_type(typeof(x)), dx))
+    dy_righttype = dy isa tangent_type(Y) ? dy : copyto!!(prep.dy_righttype, dy)
+    dx_righttype = set_to_zero!!(prep.dx_righttype)
```
willtebbutt, line 53: I'm a little confused as to how this function ever gets the right answer. How does `tx` wind up in the `__value_and_pullback!!` call? (which presumably it must do somehow if it's to get the right answer)

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
willtebbutt: When you say an error, is it a numerical error, or does something fail to run?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
gdalle: There's no `dx`, only `dy` for the pullback. And the idea is that I store `dy` in `dy_righttype`, avoiding another allocation if the type is already right

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
gdalle: It's a numerical error. Things run properly but presumably the pullback for the second tangent is wrong, whereas the pullback for the first tangent was right. So somewhere, something fails to reset between both calls

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
willtebbutt: Sorry, I'm a little confused -- what's the purpose of the `tx` argument at the top of this method?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
willtebbutt: Hmmm yeah, that is weird.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
gdalle: `tx` and `ty` are tuples of tangents. For some backends (ForwardDiff, Enzyme) I can handle them all at once. For most backends though, I only have native implementations when `t isa NTuple{1}`, and the generic implementation `map`s over the contents of the tuple. In that case, I often define `dx = only(tx)` or `dy = only(ty)`.
In this case, the pullback is calculated with seed `ty` and stored into `tx`.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
willtebbutt: Ah, I think I see what's going on. We directly return `dx_righttype`, rather than writing it out to some other buffer. Since `prep.dx_righttype` and `tangent(fdata(dx_righttype), new_dx)` can alias each other, there's memory sharing going on between each iteration.

I would imagine the the results you're seeing if you request the gradients in batch mode is that they will all be the same. Moreover, assuming you're iterating through the tangents in order, they'll all have the correct gradient for the last tangent in the batch. Does that make sense?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
willtebbutt: Ohhh sorry, I didn't see the relationship between `tx` and `dx`. This makes sense. Thanks for clarifying.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
gdalle: Oh yes, classic mistake of returning whatever is in the preparation object instead of making a copy. I'll fix it and merge

willtebbutt: Do you have the option to do prepration which is specific to batch mode? If so, you could pre-allocate buffers for each element of the batch.

gdalle: Yes, I have opened https://github.com/gdalle/DifferentiationInterface.jl/issues/518 to keep track

PR #517, opened 2024-09-30
Fix package name printing

### LINKED ISSUE #516: Package names are printed wrong when DI or ADTypes is only `import`ed
```julia
julia> import DifferentiationInterface

julia> DifferentiationInterface.package_name(DifferentiationInterface.AutoForwardDiff())
"pes.AutoForwardDiff"

julia> using DifferentiationInterface

julia> DifferentiationInterface.package_name(DifferentiationInterface.AutoForwardDiff())
"ForwardDiff"
```

gdalle:
**DI source**

- Fix package name parsing from `ADTypes` backend string

```julia
julia> import DifferentiationInterface

julia> DifferentiationInterface.package_name(DifferentiationInterface.AutoForwardDiff())
"ForwardDiff"

julia> using DifferentiationInterface

julia> DifferentiationInterface.package_name(DifferentiationInterface.AutoForwardDiff())
"ForwardDiff"
```

### DIFF
--- a/DifferentiationInterface/src/utils/printing.jl
+++ b/DifferentiationInterface/src/utils/printing.jl
@@ -1,8 +1,10 @@
 function package_name(b::AbstractADType)
     s = string(b)
+    s = chopprefix(s, "ADTypes.")
+    s = chopprefix(s, "Auto")
     k = findfirst('(', s)
     isnothing(k) && throw(ArgumentError("Cannot parse backend into package"))
-    return s[5:(k - 1)]
+    return s[begin:(k - 1)]
 end
 
 function package_name(b::SecondOrder)


DifferentiationInterface/src/utils/printing.jl
```diff
@@ -1,8 +1,10 @@
 function package_name(b::AbstractADType)
     s = string(b)
+    s = chopprefix(s, "ADTypes.")
+    s = chopprefix(s, "Auto")
```
avik-pal, line 4: a simpler version would be `nameof(typeof(b))[5:(k - 1)]`

gdalle: good point, if I ever run into another similar issue I'll remember it

PR #535, opened 2024-10-03
Improve Jacobian and Hessian preparation

### LINKED ISSUE #534: Type stability of preparation
Improve type stability of preparation despite choosing chunk size

gdalle:
**Versions**

- Bump DI to v0.6.5

**DI source**

- Encode batch size from `pick_batchsize` with `Val(B)` instead of `B`
- Pass `Val(B)` to the Jacobian and Hessian preparators with a function barrier. Do the same for sparse Jacobians and Hessians.
- Avoid slicing the whole Jacobian/Hessian when the last block is too big, only slice the last block

**DI extensions**

- Enzyme: set batch size to `Val(16)`
- ForwardDiff: set batch size to `Val(C)` if pre-selected in `AutoForwardDiff`. I tried to do the same for `Chunk{C}` but it is forbidden to have chunks larger than the input size.

**DIT source**

- Add a secret option `preparation_type_stability` to `DIT.test_differentiation` which runs `JET.@test_opt` on the preparation operator

**DI and DIT tests**

- Test `preparation_type_stability` for trivial dense and sparse backends


DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/utils.jl
```diff
@@ -1,5 +1,5 @@
 choose_chunk(::AutoForwardDiff{nothing}, x) = Chunk(x)
-choose_chunk(::AutoForwardDiff{C}, x) where {C} = Chunk{min(length(x), C)}()
+choose_chunk(::AutoForwardDiff{C}, x) where {C} = Chunk{C}()
```
gdalle: this allows ForwardDiff preparation to be type-stable

PR #580, opened 2024-10-15
Add allocation benchmarks on StaticArrays

gdalle:
**DI extensions**

- ForwardDiff: Don't reuse preallocated vector of `Dual`s in `pushforward` for out-of-place functions

> [!WARNING]
> This would cause loss of performance in other cases. The right solution would be to distinguish based on whether or not `x` allows `setindex`, but I don't want ArrayInterface.jl as a dependency.

**DI tests**

- ForwardDiff: Start testing absence of allocations on static scenarios


gdalle:
Closing because outdated, follow up in #616

PR #589, opened 2024-10-16
Apply new version of JuliaFormatter

gdalle:
Format all code with JuliaFormatter v2.0, now based on JuliaSyntax

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl
@@ -10,11 +10,11 @@ to_val(::DI.BatchSizeSettings{B}) where {B} = Val(B)
 
 ## Annotations
 
-function get_f_and_df(f::F, ::AutoEnzyme{M,Nothing}, ::Val{B}=Val(1)) where {F,M,B}
+function get_f_and_df(f::F, ::AutoEnzyme{M,Nothing}; (::Val{B})=Val(1)) where {F,M,B}
     return f
 end
 
-function get_f_and_df(f::F, ::AutoEnzyme{M,<:Const}, ::Val{B}=Val(1)) where {F,M,B}
+function get_f_and_df(f::F, ::AutoEnzyme{M,<:Const}; (::Val{B})=Val(1)) where {F,M,B}
     return Const(f)
 end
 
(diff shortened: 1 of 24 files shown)


DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl
```diff
@@ -10,11 +10,11 @@ to_val(::DI.BatchSizeSettings{B}) where {B} = Val(B)
 
 ## Annotations
 
-function get_f_and_df(f::F, ::AutoEnzyme{M,Nothing}, ::Val{B}=Val(1)) where {F,M,B}
+function get_f_and_df(f::F, ::AutoEnzyme{M,Nothing}; (::Val{B})=Val(1)) where {F,M,B}
```
gdalle, line 13: this automatic formatting changes the semantics, turning a positional argument into a keyword argument

domluna: this could be disabled with separate_kwargs_with_semicolon=false

I think this is the behaviour one would want for the option though, no? It seems if it didn't make this transformation previously it would be a bug.

domluna: or is it such that ::Val{B}=Val(1) is not considered a keyword argument but val::Val{B}=Val(1) is?

gdalle: No, previously it wasn't a bug. It was an unnamed positional argument, for which I'm only interested in the type (think of it as a Holy trait).
Besides, @MilesCranmer seems to say that the same behavior can be observed for named positional arguments, which is also problematic.
Named or unnamed, positional arguments can also have default values, as long as they come at the end of the list of positional arguments and before the semicolon. They don't need to be keyword arguments to have default values.

MilesCranmer: Yes. Here’s an example in my codebase:

```diff
function check_constraints(
     tree::AbstractExpressionNode,
     options::AbstractOptions,
-    maxsize::Int,
+    maxsize::Int;
     cursize::Union{Int,Nothing}=nothing,
 )::Bool
 ```

This means that doing a function call `check_constraints(tree, options, maxsize, cursize)` will now throw an error when previously it worked.

PR #604, opened 2024-11-03
Catch Zygote `nothing` returns and give an informative error message

ChrisRackauckas:
This is much more informative than `iterate(::Nothing)` not defined.


gdalle (changes requested):
This is not a bad idea! I would suggest

- throwing a custom error instead of the generic `ErrorException`, something like `ZygoteNothingError`
- doing the same for other operators (unless there are some which cannot return `nothing`?)
- adding tests (I'll take care of that last part)

gdalle:
Superseded by #667

PR #620, opened 2024-11-10
fix: respect array type in wrong-mode pushforward/pullback

gdalle:
**DI source**

- In `pushforward`/`pullback`, give `x`/`y` as a first argument to the `map` over cartesian indices. This allows array types (e.g. static or GPU) to be preserved.


gdalle:
The issue at the moment is with GPUArrays because `CartesianIndices(x)` is not a GPU array

gdalle:
Superseded by #974

PR #624, opened 2024-11-15
fix Diffractor compat

oscardssmith:
This was requiring an old version of Diffractor


gdalle:
Good catch but it was done on purpose, the following versions of Diffractor broke DI and no one had time to fix it. You can see it because the corresponding test is failing: https://github.com/JuliaDiff/DifferentiationInterface.jl/actions/runs/11860118951/job/33054648333?pr=624.

Related:

- https://github.com/JuliaDiff/Diffractor.jl/issues/290

gdalle:
Is it okay if I close this for the time being?

Issue #504, opened 2024-09-25
Adding TaylorDiff.jl


tansongchen:
[TaylorDiff.jl](https://github.com/JuliaDiff/TaylorDiff.jl) is a package to compute higher-order directional derivative with Taylor mode efficiently. I am interested in integrating it intuo DifferentiationInterface.jl, could you briefly describe where I should look into?

gdalle:
You can check out the [dev guide](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/dev/dev_guide/) in the documentation. There is also an ongoing integration attempt for GTPSA in #329, although it is not exactly up to date with the latest v0.6 changes.
Since I wrote all the previous AD bindings myself, it might be kind of hard to understand how everything fits together, so don't hesitate to ask me!

PR #635, opened 2024-11-21
[WIP] Add TaylorDiff

tansongchen:
will solve #504


gdalle:
Since it's been a year and both packages have evolved significantly, I'm gonna go ahead and close this. Feel free to reopen if still relevant, although higher-order differentiation is not the focus of DI.

PR #645, opened 2024-11-28
Mooncake Upgrades

willtebbutt:
In response to a couple of things, I've decided to add much better tools for caching various things in Mooncake. As a result, we can substantially simplify Mooncake's implementation in DifferentiationInterface.

I've tested this locally, so I'm hoping it still works...


willtebbutt:
(failures because the right version of Mooncake is not yet available, because I messed up doing releases)

willtebbutt:
(codecov is lying -- the line it says is uncovered is definitely covered)

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl
```diff
@@ -1,65 +1,49 @@
...
     f, backend::AutoMooncake, x, ty::NTuple, contexts::Vararg{Context,C}
 ) where {C}
-    y = f(x, map(unwrap, contexts)...)
-    config = get_config(backend)
-    rrule = build_rrule(
-        get_interpreter(),
-        Tuple{typeof(f),typeof(x),typeof.(map(unwrap, contexts))...};
-        debug_mode=config.debug_mode,
-        silence_debug_messages=config.silence_debug_messages,
+    config = backend.config === nothing ? Mooncake.Config() : backend.config
```
gdalle:
```suggestion
    config = get_config(backend)
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
```diff
@@ -1,32 +1,19 @@
...
-    )
-    df! = zero_tangent(f!)
-    y_copy = copy(y)
-    dx_righttype = zero_tangent(x)
-    dy_righttype = zero_tangent(y)
+    target_function = function (f!, y, x, contexts...)
+        f!(y, x, contexts...)
+        return y
+    end
+    cache = prepare_pullback_cache(target_function, f!, y, x, map(unwrap, contexts)...)
```
gdalle: is there a reason why don't use the config here?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
```diff
@@ -1,32 +1,19 @@
...
-        get_interpreter(),
-        Tuple{typeof(f!),typeof(y),typeof(x),typeof.(map(unwrap, contexts))...};
-        debug_mode=config.debug_mode,
-        silence_debug_messages=config.silence_debug_messages,
-    )
-    df! = zero_tangent(f!)
-    y_copy = copy(y)
-    dx_righttype = zero_tangent(x)
-    dy_righttype = zero_tangent(y)
+    target_function = function (f!, y, x, contexts...)
```
gdalle, line 10: will this new closure slow things down?

gdalle (commented):
Thanks! This new cache thingy isn't documented anywhere in Mooncake.jl, is it the new name for rules? Is it an internal or part of the API?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
willtebbutt: I don't believe so -- Mooncake generally has no problems with closures.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl
willtebbutt: No, that is a mistake on my part.

willtebbutt: Resolved by latest commit

willtebbutt:
> This new cache thingy isn't documented anywhere in Mooncake.jl, is it the new name for rules?

It's a wrapper around a rule, and all of the various tangents that you need in order to avoid allocations. Unlike `build_rrule`, which just used static information, the `cache` makes use of dynamic information (e.g. array sizes) which it uses to make life easier.

> Is it an internal or part of the API?

It's internal for now. I plan to make it available publicly in short order, but want to make sure it's doing everything as I intend it before telling users about it.

Issue #646, opened 2024-11-29
Complex number support


ErikQQY:
Find this while implementing https://github.com/SciML/BoundaryValueDiffEq.jl/pull/258

This MWE is just proof of the idea and may not be meaningful, but I think it can still prove that the current DI lacks support for sparse Jacobian of complex numbers.

```julia
using ForwardDiff, DifferentiationInterface, SparseMatrixColorings, ADTypes
const DI = DifferentiationInterface
#backend = AutoForwardDiff() # this works
backend = AutoSparse(
    AutoForwardDiff(),
    sparsity_detector = ADTypes.KnownJacobianSparsityDetector(ones(3, 3)),
    coloring_algorithm = ConstantColoringAlgorithm(ones(3, 3), ones(Int64, 3))
)
u0 = [1.0; 0.0; 0.0] .+1im
jac_cache = DI.prepare_jacobian(nothing, similar(u0), backend, u0)
```

stack trace

```julia
ERROR: MethodError: no method matching DifferentiationInterfaceSparseMatrixColoringsExt.PushforwardSparseJacobianPrep(::DifferentiationInterface.BatchSizeSettings{…}, ::SparseMatrixColorings.ColumnColoringResult{…}, ::Matrix{…}, ::Vector{…}, ::Vector{…}, ::DifferentiationInterfaceForwardDiffExt.ForwardDiffTwoArgPushforwardPrep{…})

Closest candidates are:
  DifferentiationInterfaceSparseMatrixColoringsExt.PushforwardSparseJacobianPrep(::BS, ::C, ::M, ::S, ::R, ::E) where {BS<:DifferentiationInterface.BatchSizeSettings, C<:(AbstractColoringResult{:nonsymmetric, :column}), M<:(AbstractMatrix{<:Real}), S<:(AbstractVector{<:Tuple{Vararg{T, N}} where {N, T}}), R<:(AbstractVector{<:Tuple{Vararg{T, N}} where {N, T}}), E<:DifferentiationInterface.PushforwardPrep}
   @ DifferentiationInterfaceSparseMatrixColoringsExt ~/.julia/packages/DifferentiationInterface/bulUW/ext/DifferentiationInterfaceSparseMatrixColoringsExt/jacobian.jl:11

Stacktrace:
[...]

PR #647, opened 2024-11-30
Relax type annotations in sparse prep

ErikQQY:
Part of #646 

Relax type annotations of the compressed matrix from `AbstractMatrix{<:Real}` to `AbstractMatrix{<:Number}` in `PushforwardSparseJacobianPrep`, `PullbackSparseJacobianPrep`, `SparseHessianPrep` and `MixedModeSparseJacobianPrep` to support sparse Jacobian/Hessian preparation for complex numbers.

The test case for this PR is adapted from MWE,  so need some advice on this.

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceSparseMatrixColoringsExt/hessian.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceSparseMatrixColoringsExt/hessian.jl
@@ -1,7 +1,7 @@
 struct SparseHessianPrep{
     BS<:BatchSizeSettings,
     C<:AbstractColoringResult{:symmetric,:column},
-    M<:AbstractMatrix{<:Real},
+    M<:AbstractMatrix{<:Number},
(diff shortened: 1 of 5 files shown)


gdalle:
Thanks! I can take over for the tests

gdalle:
I assume you were using complex numbers with FiniteDiff? I tried with ForwardDiff but it won't even build the `Dual`s

gdalle (approved):

ErikQQY:
Yeah, it should be FiniteDiff, very sorry I may not have specified this clearly. When handling complex numbers, in my specific use case, according to the input `x` type(ComplexF64, etc.) in `prepare_jacobian`, FiniteDiff backend should be used. But the general `prepare_jacobian` API failed to handle this no matter which backend is being used.

ErikQQY:
Can we cut a release for this patch?

gdalle:
Here you go: https://github.com/JuliaRegistries/General/pull/120422

PR #673, opened 2025-01-06
Fix typo for `gradient` in docs

mattsignorelli:
This fixes a minor typo in the docs for `gradient`, as it must take an array and not `Any`, e.g.
```julia
julia> gradient(x->sin(x), AutoForwardDiff(), 1)
ERROR: DimensionMismatch: gradient(f, x) expects that x is an array. Perhaps you meant derivative(f, x)?
Stacktrace:
 [1] gradient(f::Function, x::Int64)
   @ ForwardDiff ~/.julia/packages/ForwardDiff/UBbGT/src/gradient.jl:46
 [2] gradient(::var"#1#2", ::AutoForwardDiff{nothing, Nothing}, ::Int64)
   @ DifferentiationInterfaceForwardDiffExt ~/.julia/dev/DifferentiationInterface/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl:295
 [3] top-level scope
   @ REPL[5]:1
```


gdalle:
This is actually not a typo, for some backends `gradient` should be able to take `Any` as input. It just so happens that ForwardDiff wants an array, but for instance Zygote would handle that perfectly :)

PR #693, opened 2025-01-21
Add relstep and absstep parameters

### LINKED ISSUE #688: Expose step size parameters for AutoFiniteDiff
When using `AutoFiniteDiff`, one should be able to change the `absstep` and `relstep` parameters as described in the [FiniteDiff.jl documentation](https://docs.sciml.ai/FiniteDiff/stable/api/#Jacobians). This seems to be impossible with the current DifferentiationInterface.jl API.

From this question on the Julia discourse: https://discourse.julialang.org/t/changing-the-relstep-of-autofinitediff-in-an-optimizationfunction/125004

1-Bart-1:
Fixes #688 

@gdalle Should I add any specific tests for relstep and absstep?


DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -43,6 +43,8 @@ end
 
 struct FiniteDiffOneArgDerivativePrep{C} <: DI.DerivativePrep
     cache::C
+    relstep
+    absstep
```
gdalle: these are already stored in the backend object, so we don't need them in the preparation

gdalle (changes requested):

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
1-Bart-1: But the backend is immutable. If I remove relstep and absstep from prep these lines would be repeated each time relstep and absstep are used:
```julia
relstep =
        isnothing(backend.relstep) ? default_relstep(fdtype, eltype(x)) : backend.relstep
absstep = isnothing(backend.absstep) ? relstep : backend.relstep
```
If you prefer this I can of course change this.

gdalle: Okay, it makes sense to put those in the `prep` object then, sorry. Two things to be mindful of:
- there is one preparation type per operator, you only modified the first one (for `derivative`)
- the fields need to have a parametric type, right now in your version they have no annotation, which is bad for type-stability because it is equivalent to `step::Any`

gdalle:
If you want tests to pass, you will need to run `JuliaFormatter.format`. Otherwise the `Core/Internals` group will fail first and you won't see the results of other test groups

1-Bart-1:
I ran the formatting now, but I don't know why the tests are failing now. Somehow, prepare_derivative is getting called with a prep stuct instead of a normal array for y?

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
+    absstep = isnothing(backend.absstep) ? relstep : backend.relstep
+    return FiniteDiffOneArgDerivativePrep(cache, relstep, absstep)
 end
 
 ### Scalar to scalar
 
 function DI.derivative(
     f,
-    ::FiniteDiffOneArgDerivativePrep{Nothing},
+    ::FiniteDiffOneArgDerivativePrep{Nothing,Nothing,Nothing},
```
gdalle: this dispatch is not possible because you make sure in the preparation that the steps will not be `nothing`
```suggestion
    ::FiniteDiffOneArgDerivativePrep{Nothing},
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
 ) where {C}
     fc = DI.with_contexts(f, contexts...)
-    return finite_difference_derivative(fc, x, fdtype(backend))
+    relstep = if isnothing(backend.relstep)
+        default_relstep(fdtype(backend), eltype(x))
+    else
+        backend.relstep
+    end
+    absstep = isnothing(backend.absstep) ? relstep : backend.relstep
+    return finite_difference_derivative(fc, x, fdtype(backend); relstep, absstep)
```
gdalle:
```suggestion
    (; relstep, absstep) = prep
    return finite_difference_derivative(fc, x, fdtype(backend); relstep, absstep)
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
 end
 
 function DI.value_and_derivative(
     f,
-    ::FiniteDiffOneArgDerivativePrep{Nothing},
+    ::FiniteDiffOneArgDerivativePrep{Nothing,Nothing,Nothing},
```
gdalle:
```suggestion
    ::FiniteDiffOneArgDerivativePrep{Nothing},
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
     y = fc(x)
-    return y, finite_difference_derivative(fc, x, fdtype(backend), eltype(y), y)
+    relstep = if isnothing(backend.relstep)
+        default_relstep(fdtype(backend), eltype(x))
+    else
+        backend.relstep
+    end
+    absstep = isnothing(backend.absstep) ? relstep : backend.relstep
+    return y,
+    finite_difference_derivative(fc, x, fdtype(backend), eltype(y), y; relstep, absstep)
```
gdalle:
```suggestion
    (; relstep, absstep) = prep
    return y,
    finite_difference_derivative(fc, x, fdtype(backend), eltype(y), y; relstep, absstep)
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
 end
 
 ### Scalar to array
 
 function DI.derivative(
     f,
-    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
+    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache,<:Number,<:Number},
```
gdalle: again this is not necessary because you already make sure they are numbers ;)
```suggestion
    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
+    return finite_difference_gradient(
+        fc, x, prep.cache; relstep=prep.relstep, absstep=prep.absstep
+    )
 end
 
 function DI.derivative!(
     f,
     der,
-    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
+    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache,<:Number,<:Number},
```
gdalle:
```suggestion
    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
-    return finite_difference_gradient!(der, fc, x, prep.cache)
+    return finite_difference_gradient!(
+        der, fc, x, prep.cache; relstep=prep.relstep, absstep=prep.absstep
+    )
 end
 
 function DI.value_and_derivative(
     f,
-    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
+    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache,<:Number,<:Number},
```
gdalle:
```suggestion
    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -56,87 +78,118 @@ function DI.prepare_derivative(
...
+    finite_difference_gradient(
+        fc, x, prep.cache; relstep=prep.relstep, absstep=prep.absstep
+    )
 end
 
 function DI.value_and_derivative!(
     f,
     der,
-    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
+    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache,<:Number,<:Number},
```
gdalle:
```suggestion
    prep::FiniteDiffOneArgDerivativePrep{<:GradientCache},
```

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/twoarg.jl
```diff
@@ -33,14 +45,22 @@ end
 
 struct FiniteDiffTwoArgDerivativePrep{C} <: DI.DerivativePrep
     cache::C
+    relstep
+    absstep
 end
```
gdalle:
```suggestion
struct FiniteDiffTwoArgDerivativePrep{C,R,A} <: DI.DerivativePrep
    cache::C
    relstep::R
    absstep::A
end
```
when we forget type annotations, the types are designated as `::Any` and this causes many instabilities down the line

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
gdalle: it also explains the error you encountered in CI, where the dispatch chain was broken and DI mistook the `prep` object in `derivative(f, prep, backend, x)` for a `y` object in `derivative(f!, y, backend, x)`

gdalle (changes requested):
Thanks a lot for giving this a shot! I made a few remarks, feel free to reach out if anything is unclear

gdalle:
I recommend you read these comments from the "Files changed" tab and use the "Add suggestion to batch" to apply them if you agree. It is much easier than 10 individual commits, one for each line

1-Bart-1:
Yes, found out about that a bit too late :laughing:

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -168,7 +216,8 @@ function DI.value_and_gradient(
     contexts::Vararg{DI.Context,C},
 ) where {C}
     fc = DI.with_contexts(f, contexts...)
-    return fc(x), finite_difference_gradient(fc, x, prep.cache)
+    return fc(x), (; relstep, absstep) = prep
+    return finite_difference_gradient(fc, x, prep.cache; relstep, absstep)
```
gdalle:
```suggestion
    (; relstep, absstep) = prep
    return fc(x), finite_difference_gradient(fc, x, prep.cache; relstep, absstep)
```
I seem to have screwed up here

1-Bart-1:
The tests are not passing. Could this be because functions are called with the same prep, but with backends with different fdtypes? If that is the case I was wrong about putting the relstep and absstep in the prep...

gdalle:
I think it might be due to the choice of `relstep`, which depends on `fdtype`... OR `fdjtype` (for Jacobian) OR `fdhtype` (for Hessian). The errors are probably due to the copypasting without operator-specific adaptation

1-Bart-1:
Nice, the FiniteDiff test passed!

gdalle:
Thanks for your efforts! I'll wait until https://github.com/SciML/ADTypes.jl/pull/104 is merged, add a few more tweaks and then this will be good to go. Any idea why all lines are hit even though we didn't add specific tests where the `step`s in the backend differ from `nothing`?

1-Bart-1:
No problem! No, I dont know… I expected we would need to add extra tests.On 23 Jan 2025, at 21:21, Guillaume Dalle ***@***.***> wrote:﻿
Thanks for your efforts! I'll wait until SciML/ADTypes.jl#104 is merged, add a few more tweaks and then this will be good to go. Any idea why all lines are hit even though we didn't add specific tests where the steps in the backend differ from nothing?

gdalle (approved):

PR #705, opened 2025-01-28
feat: use `JVPCache` for FiniteDiff pushforwards

gdalle:
- For array-to-array pushforwards with FiniteDiff, enable preparation thanks to the new `JVPCache`
- Add benchmarks proving that sparse Jacobians are now alloc-free for ForwardDiff and FiniteDiff

Related:

- https://github.com/JuliaDiff/FiniteDiff.jl/pull/191

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.6.37"
+version = "0.6.38"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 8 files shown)


oscardssmith:
Is this sufficient for sparse finitediff jacobians to be supported?

gdalle:
Yes, but we'll have to benchmark performance to make sure that they are non-allocating

oscardssmith:
The in place is fully non-allocating.
```
julia> g(dx,x) = dx.=x*2
g (generic function with 1 method)

julia> cache = FiniteDiff.JVPCache(rand(3), rand(3), Val(:forward))
FiniteDiff.JVPCache{Vector{Float64}, Vector{Float64}, Val{:forward}()}([0.3185378518133124, 0.9795057009072153, 0.007265694202993944], [0.25969842682138, 0.5029073571918922, 0.41967454290961415])

julia> @btime FiniteDiff.finite_difference_jvp!($(ones(3)), g, $(ones(3)), $(ones(3)), $cache)
  25.392 ns (0 allocations: 0 bytes)
```

gdalle:
Nice! I'll add a non-allocation test to this PR to prevent regressions

PR #714, opened 2025-02-01
fix: revert catch of nothing pullbacks with Zygote

### LINKED ISSUE #713: Interpretation of `nothing` from Zygote backend likely incorrect?
I think the solution in #667 (fixing #662) likely isn't correct, because now it means functions that do not explicitly depend on the input will hit errors.

While `nothing` is used by Zygote to indicate that a rule has not been implemented, it might ALSO simply mean that there's no dependence. Thus I think throwing an error on **every** `nothing` is wrong. For example, this means we can't differentiate `f(x) = 0.0` in Zygote, though other backends work:

```julia
julia> import Zygote, ForwardDiff; using DifferentiationInterface

julia> DifferentiationInterface.gradient(x -> 0.0, AutoForwardDiff(), [0.0])
1-element Vector{Float64}:
 0.0

julia> DifferentiationInterface.gradient(x -> 0.0, AutoZygote(), [0.0])
ERROR: Zygote failed to differentiate function `#11` with argument types `(Vector{Float64},)` (the pullback returned `nothing`).
Stacktrace:
 [1] check_nothing(::Nothing, f::Function, x::Vector{Float64}, contexts::Tuple{})
   @ DifferentiationInterfaceZygoteExt ~/.julia/packages/DifferentiationInterface/TFo7D/ext/DifferentiationInterfaceZygoteExt/DifferentiationInterfaceZygoteExt.jl:31
```

I think the best behavior is to just return `nothing` - it's not perfect, but it will avoid flagging cases like this. Even differentiating the function `Base.zero` is now an error, because of the fact that

```julia
julia> Zygote.gradient(zero, 1.0)
(nothing,)
```

Ran into this for https://github.com/MilesCranmer/SymbolicRegression.jl/pull/404

gdalle:
fix #713

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.6.38"
+version = "0.6.39"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/DifferentiationInterface/ext/DifferentiationInterfaceZygoteExt/DifferentiationInterfaceZygoteExt.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceZygoteExt/DifferentiationInterfaceZygoteExt.jl
@@ -13,24 +13,6 @@ using Zygote:
     withgradient,
     withjacobian
 
-struct ZygoteNothingError <: Exception
-    f
-    x
-    contexts
-end
-
-function Base.showerror(io::IO, e::ZygoteNothingError)
-    (; f, x, contexts) = e
-    sig = (typeof(x), map(typeof ∘ DI.unwrap, contexts)...)
-    return print(
-        io,
-        "Zygote failed to differentiate function `$f` with argument types `$sig` (the pullback returned `nothing`).",
-    )
-end
-
-check_nothing(::Nothing, f, x, contexts) = throw(ZygoteNothingError(f, x, contexts))
-check_nothing(::Any, f, x, contexts) = nothing
-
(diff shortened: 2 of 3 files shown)


MilesCranmer:
Thanks for getting to this so quickly. Do you think you can release a new version with this patch?

gdalle:
I was just wondering whether I should do anything more. Right now we're back to the old behavior with errors involving `nothing` at random locations.
Essentially when you call a DI operator which directly falls back on Zygote, you're okay. But when you call something more sophisticated which doesn't have a direct Zygote equivalent, a `nothing` will appear _in the middle of the code_ and then there's no telling where things will blow up, but importantly, they will blow up in a way that you can't really `try/catch` yourself for lack of a predictable exception type.
I thought about introducing a custom exception like you said, but where do we draw the line? What if there's an array for which some elements are `nothing` and some are not (as you [demonstrated](https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/713#issuecomment-2628537013)? Should I recursively dive into substructures to check for nothingness? These questions made me think the default behavior of letting things error if the rule is not defined is the right choice. After all this is a Zygote problem, not a me problem?

MilesCranmer:
I think this is fundamentally a Zygote problem; there's nothing that can be done in DifferentiationInterface.jl to solve it. It needs to be fixed within Zygote.jl itself by creating new return types for each scenario. Leaving it as `nothing` as-is is not ideal, but it's basically the _least bad approach_. Trying to handle `nothing` explicitly is too dangerous, because (1) sometimes `nothing` is _real_ (a field of a struct could just literally be nothing), (2) sometimes `nothing` indicates no dependence, and (3) sometimes `nothing` indicates a lack of a rule.

No way to identify which of these three branches was hit without changing Zygote itself. In a perfect world there would be a `NoTangent()` and `UndefinedTangent()` return value from Zygote to help disentangle these.

gdalle:
alright then, I'll release as-is

PR #715, opened 2025-02-03
Update citation files with preprint 

adrhill:


DifferentiationInterface/README.md
```diff
@@ -90,4 +90,35 @@ To improve your performance by up to several orders of magnitude compared to thi
 
 ## Citation
 
-Please cite both DifferentiationInterface.jl and its inspiration [AbstractDifferentiation.jl](https://github.com/JuliaDiff/AbstractDifferentiation.jl), using the provided `CITATION.cff` file.
+Please cite our preprint [*Sparser, Better, Faster, Stronger: Efficient Automatic Differentiation for Sparse Jacobians and Hessians*](https://arxiv.org/abs/2501.17737), the [DifferentiationInterface.jl software](https://zenodo.org/records/11092033) and its inspiration [AbstractDifferentiation.jl](https://github.com/JuliaDiff/AbstractDifferentiation.jl), using the provided `CITATION.cff` file:
```
gdalle: Perhaps specify "CFF or BibTex"?
Also, researchers only need to cite the software itself (in addition to the papers) if they actually use it. But if they only refer to the ideas, they only need to cite the papers.

gdalle (changes requested):

DifferentiationInterface/README.md
gdalle: So we could say something like:

Please cite our preprint and the AbstractDiff preprint whenever you refer to this package or the ideas it contains. In addition, if you actually rely on our code, please _also_ cite the precise Zenodo DOI of the version you used.

gdalle (approved):
thanks!

PR #725, opened 2025-02-10
test: add Mooncake to Lux and Flux tests

gdalle:
At the moment, the Lux tests are still based on ComponentArrays, we should probably change that to handle native parameter gradients


gdalle:
Closing because it requires https://github.com/compintell/Mooncake.jl/issues/472

PR #729, opened 2025-02-14
perf: avoid double function call in ReverseDiff `value_and_gradient`

gdalle:
Personal musings: The behavior of ReverseDiff is very confusing, even on `ImmutableDiffResult`. I encountered a Heisenbug which seems to depend on the compilation path (disappears when I add a `print` statement), where `value_and_gradient!` suddenly becomes incorrect when called inside `value_gradient_and_hessian!`. But only for a compiled tape. And the `:gradient` tests for `value_and_gradient!` still pass. What a mess.

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.6.40"
+version = "0.6.41"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/DifferentiationInterface/ext/DifferentiationInterfaceReverseDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceReverseDiffExt/onearg.jl
@@ -87,23 +87,27 @@ end
 function DI.value_and_gradient!(
     f, grad, prep::ReverseDiffGradientPrep, ::AutoReverseDiff{compile}, x
 ) where {compile}
-    y = f(x)  # TODO: ReverseDiff#251
-    result = DiffResult(y, (grad,))
+    # DiffResult doesn't work because of ReverseDiff#251
+    result = MutableDiffResult(first(x), (grad,))
     if compile
         result = gradient!(result, prep.tape, x)
     else
         result = gradient!(result, f, x, prep.config)
     end
-    y = DR.value(result)
-    grad === DR.gradient(result) || copyto!(grad, DR.gradient(result))
-    return y, grad
+    return DR.value(result), DR.gradient(result)
 end
 
 function DI.value_and_gradient(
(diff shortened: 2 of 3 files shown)


gdalle:
So, in the current state of the PR, `value_and_gradient!` works but not alongside a Hessian computation... unless we print something? I'm going crazy.

```julia
julia> DifferentiationInterface.value_gradient_and_hessian!(
           f,
           zeros(2),
           zeros(2, 2),
           SecondOrder(AutoFiniteDiff(), AutoReverseDiff(; compile=true)),
           ones(2),
       )
(0.0, [2.0, 2.0], [2.0 0.0; 0.0 2.0])

julia> DifferentiationInterface.value_gradient_and_hessian!(
           f,
           zeros(2),
           zeros(2, 2),
           SecondOrder(AutoFiniteDiff(), AutoReverseDiff(; compile=true)),
           ones(2),
       )
(0.0, [2.0, 2.0], [2.0 0.0; 0.0 2.0])

julia> using DifferentiationInterface, FiniteDiff, ReverseDiff

julia> DifferentiationInterface.value_and_gradient!(
           f,
           zeros(2),
           AutoReverseDiff(; compile=true),
           ones(2),
       )
(2.0, [2.0, 2.0])

julia> DifferentiationInterface.value_gradient_and_hessian!(
           f,
           zeros(2),
           zeros(2, 2),
           SecondOrder(AutoFiniteDiff(), AutoReverseDiff(; compile=true)),
           ones(2),
       )
(0.0, [2.0, 2.0], [2.0 0.0; 0.0 2.0])

julia> # now we just add a log

julia> DifferentiationInterface.value_and_gradient!(
           f,
           zeros(2),
           AutoReverseDiff(; compile=true),
           ones(2),
       )
[ Info: I'm here
(2.0, [2.0, 2.0])

julia> DifferentiationInterface.value_gradient_and_hessian!(
           f,
           zeros(2),
           zeros(2, 2),
           SecondOrder(AutoFiniteDiff(), AutoReverseDiff(; compile=true)),
           ones(2),
       )
[ Info: I'm here
(2.0, [2.0, 2.0], [2.0 0.0; 0.0 2.0])
```

gdalle:
I further boiled it down to whether I do `y, grad` or `y, _` in the assignment. Still have no clue what's going on.

EDIT: pure-ReverseDiff example available at https://github.com/JuliaDiff/ReverseDiff.jl/issues/269

```julia
using DifferentiationInterface  # version from this branch
import DifferentiationInterface as DI
using ReverseDiff: ReverseDiff

backend = AutoReverseDiff(; compile=true)
f(x) = sum(abs2, x)
x = ones(2)

prep = prepare_gradient(f, backend, zero(x))

function value_and_gradient_nested!(f, grad, prep, backend, x)
    y, _ = value_and_gradient!(f, grad, prep, backend, x)
    return y, grad
end
```

```julia
julia> value_and_gradient!(f, zeros(2), prep, backend, x)
(2.0, [2.0, 2.0])

julia> value_and_gradient_nested!(f, zeros(2), prep, backend, x)  # wrong
(0.0, [2.0, 2.0])
```

This behavior only happens with the compiled tape mode of ReverseDiff. I think it might be because the compiler struggles to figure out that ReverseDiff mutates the value of `y` in the `MutableDiffResult`?

PR #741, opened 2025-03-14
perf: preallocate `Cache` during preparation for ForwardDiff

gdalle:
- Add a `contexts_dual` field to most preparation objects in the ForwardDiff extension
- Initialize this field with a tuple of the same size as `contexts`, except that
  - non-`Cache` contexts are replaced by `nothing`
  - `Cache` contexts are replaced by `similar(c, Dual{...})`
- During prepared execution, take the `Cache` contexts from `prep.contexts_dual`
- Create `Config`s with `f=nothing` because the `tag` captures the relevant type
- Only take the no-preparation shortcuts when there are no `Cache` contexts
- Temporarily deactivate (Polyester)ForwardDiff HVP optimizations, must be reactivated before the next release

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.6.45"
+version = "0.6.46"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/DifferentiationInterfaceForwardDiffExt.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/DifferentiationInterfaceForwardDiffExt.jl
@@ -24,6 +24,7 @@ using ForwardDiff:
     jacobian,
     jacobian!,
     partials,
+    pickchunksize,
     value
 
 DI.check_available(::AutoForwardDiff) = true
@@ -31,7 +32,7 @@ DI.check_available(::AutoForwardDiff) = true
 include("utils.jl")
 include("onearg.jl")
 include("twoarg.jl")
-include("secondorder.jl")
+# include("secondorder.jl")
 include("differentiate_with.jl")
 include("misc.jl")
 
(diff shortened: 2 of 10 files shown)


franckgaga:
Yes `Cache` context does not allocate at runtime with `AutoForwardDiff`:

```julia
using BenchmarkTools
import ForwardDiff
x = float.(1:8)
y = Vector{eltype(x)}(undef, length(x)-1)
cache = zeros(eltype(x), length(y))
function f!(y, x, cache)
    y2 = cache
    for i in eachindex(y)
        y[i] = (x[i+1]^2-x[i]^2)
    end
    for i in eachindex(y)
        j = length(x) - i
        y[i] += (x[j]^2-x[j+1]^2)
    end
    y2 .= y
    return y
end
dense_backend = AutoForwardDiff()
dense_prep = prepare_jacobian(f!, y, dense_backend, x, Cache(cache))
∇f = Matrix{eltype(x)}(undef, length(y), length(x))
jacobian!(f!, y, ∇f, dense_prep, dense_backend, x, Cache(cache))
display(∇f)  
@benchmark jacobian!($f!, $y, $∇f, $dense_prep, $dense_backend, $x, Cache($cache))
```
giving:
```julia
7×8 Matrix{Float64}:
 -2.0   4.0   0.0   0.0    0.0    0.0   14.0  -16.0
  0.0  -4.0   6.0   0.0    0.0   12.0  -14.0    0.0
  0.0   0.0  -6.0   8.0   10.0  -12.0    0.0    0.0
  0.0   0.0   0.0   0.0    0.0    0.0    0.0    0.0
  0.0   0.0   6.0  -8.0  -10.0   12.0    0.0    0.0
  0.0   4.0  -6.0   0.0    0.0  -12.0   14.0    0.0
  2.0  -4.0   0.0   0.0    0.0    0.0  -14.0   16.0
BenchmarkTools.Trial: 10000 samples with 907 evaluations per sample.
 Range (min … max):  108.548 ns … 567.631 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     125.157 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   135.268 ns ±  33.886 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄▅▆▇▆█▆▆▆▄▅▅▄▃▃▂▁▁                                            ▂
  █████████████████████▆▇▅▇▆▆▅▆▅▅▅▅▄▆▆▆▇▇████▇██▆▆▆▅▅▄▅▅▆▅▃▆▅▄▃ █
  109 ns        Histogram: log(frequency) by time        289 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```
Also allocation-free with `AutoFiniteDiff`.

Good work!

PR #752, opened 2025-03-19
Remove 1 backtick in excess

franckgaga:
The exemple does not render and the jldoctest is probably not executed.

Keep up the good work!


gdalle (approved):
thanks!

PR #759, opened 2025-03-27
perf: use common subexpression elimination in AutoSymbolics

### LINKED ISSUE #758: `Symbolics.build_funciton` now supports common-subexpression elimination
`build_function(...; cse=true)` can now yield enormous speedups, [the simple benchmark here](https://github.com/JuliaSymbolics/Symbolics.jl/issues/743) went from 2.490 μs to 184.236 ns for a sparse Hessian. It would be nice if this was activated in DI.

@AayushSabharwal are there any cons to enabling it by default?

baggepinnen:
closes #758


gdalle (approved):
Thanks!

PR #760, opened 2025-03-28
fix: erase cache contexts before building function in Symbolics

gdalle:
This PR fixes bad cache handling with Symbolics.
The cache variables need to be erased or otherwise replaced before the symbolic function is built. Otherwise the derivatives are deduced from the caches and not from the active variable, as seen in https://github.com/JuliaControl/ModelPredictiveControl.jl/pull/174#issuecomment-2760465744


gdalle:
@brianguenter do you have any idea why such a fix isn't necessary with FastDifferentiation?

brianguenter:
@gdalle I do not. Symbolics uses very different representations of symbolic expressions than FastDifferentiation does. Somewhere in the nitty gritty details of these differences lies the cause.

gdalle:
Alright, I guess the mystery will remain!

PR #768, opened 2025-04-01
feat: backend switching for Mooncake

AstitvaAggarwal:
Define `Mooncake.rrule!!` for `DI.DifferentiateWith`.


AstitvaAggarwal:
removed the code that piggybacks off the Chainrules wrapper. This is specifically now a Mooncake generic rule which handles backend switching.

DifferentiationInterface/Project.toml
```diff
@@ -38,7 +38,7 @@ DifferentiationInterfaceFiniteDiffExt = "FiniteDiff"
 DifferentiationInterfaceFiniteDifferencesExt = "FiniteDifferences"
 DifferentiationInterfaceForwardDiffExt = ["ForwardDiff", "DiffResults"]
 DifferentiationInterfaceGTPSAExt = "GTPSA"
-DifferentiationInterfaceMooncakeExt = "Mooncake"
+DifferentiationInterfaceMooncakeExt = ["ChainRulesCore", "Mooncake"]
```
```suggestion
DifferentiationInterfaceMooncakeExt = "Mooncake"
```
I think you're no longer piggybacking on ChainRules? Besides, even if you were, this addition would mean that users need to load ChainRulesCore in order to call Mooncake through DI, effectively breaking every downstream code that relies on `AutoMooncake`. Does that make sense?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/DifferentiationInterfaceMooncakeExt.jl
```diff
@@ -3,14 +3,21 @@ module DifferentiationInterfaceMooncakeExt
...
     value_and_gradient!!,
     value_and_pullback!!,
-    zero_tangent
+    zero_tangent,
+    @is_primitive,
+    zero_fcodual,
+    MinimalCtx,
+    NoRData
+
+using ChainRulesCore: ChainRulesCore, rrule
```
gdalle:
```suggestion
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
+@is_primitive MinimalCtx Tuple{CoDual{<:DI.DifferentiateWith},CoDual{<:AbstractArray}}
+@is_primitive MinimalCtx Tuple{CoDual{<:DI.DifferentiateWith},CoDual{<:Number}}
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, args::CoDual...)
+    primal_func = Mooncake.primal(dw)
+    primal_args = map(arg -> Mooncake.primal(arg), args)
```
gdalle:
```suggestion
    primal_x = primal(x)
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/DifferentiationInterfaceMooncakeExt.jl
```diff
@@ -3,14 +3,21 @@ module DifferentiationInterfaceMooncakeExt
...
     prepare_pullback_cache,
     tangent_type,
     value_and_gradient!!,
     value_and_pullback!!,
-    zero_tangent
+    zero_tangent,
+    @is_primitive,
+    zero_fcodual,
+    MinimalCtx,
+    NoRData
```
gdalle:
```suggestion
    NoRData,
    primal
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
+@is_primitive MinimalCtx Tuple{CoDual{<:DI.DifferentiateWith},CoDual{<:AbstractArray}}
+@is_primitive MinimalCtx Tuple{CoDual{<:DI.DifferentiateWith},CoDual{<:Number}}
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, args::CoDual...)
+    primal_func = Mooncake.primal(dw)
```
gdalle:
```suggestion
    primal_func = primal(dw)
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
+@is_primitive MinimalCtx Tuple{CoDual{<:DI.DifferentiateWith},CoDual{<:AbstractArray}}
+@is_primitive MinimalCtx Tuple{CoDual{<:DI.DifferentiateWith},CoDual{<:Number}}
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, args::CoDual...)
```
gdalle:
```suggestion
function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, x::CoDual)
```
Right now this only works for a single argument anyway, right?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
...
+    primal_func = Mooncake.primal(dw)
+    primal_args = map(arg -> Mooncake.primal(arg), args)
+
+    (; f, backend) = primal_func
+    y = f(primal_args...)
```
gdalle:
```suggestion
    y = f(primal_x)
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
...
+
+    prep_same = DI.prepare_pullback_same_point_nokwarg(
+        Val(true), f, backend, primal_args..., (y,)
+    )
```
gdalle:
```suggestion
```
Unlike Zygote, Mooncake supports mutation, which also means it cannot handle calling two pullbacks in a row without doing a forward pass in-between (as discussed in https://github.com/JuliaDiff/DifferentiationInterface.jl/discussions/678). Thus, I don't think preparation brings anything here.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
...
+
+    function pullback!!(dy)
+        tx = DI.pullback(f, prep_same, backend, primal_args, (dy,))
```
gdalle:
```suggestion
        tx = DI.pullback(f, backend, primal_args, (dy,))
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
...
+        args_rdata = map((x) -> (x, Mooncake.zero_rdata(x)), only(tx))
+        return NoRData(), args_rdata...
```
gdalle: I don't think this is correct:
- if `x isa AbstractArray` you need to increment its `FData` with `dx = only(tx)`, and it has `NoRData()`
- if `x isa Number` it has `NoFData()` and you need to return `dx` as its `RData`

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    return zero_fcodual(y), pullback!!
```
gdalle: What does `zero_fcodual` do?

gdalle (changes requested):
Thanks for this first draft!
I think there are some changes necessary, and most importantly you need to test it, first locally and then during CI (try not to run CI before having tested your changes locally, the process is very expensive since it tests a dozen different backends for like half an hour each).
For the testing, start with manual tests, and then once your code works you can add `AutoMooncake()` to [this line](https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/ba8a2fa9e7558e65655aeedbc36158e41ea5e324/DifferentiationInterface/test/Back/DifferentiateWith/test.jl#L27)

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: it returns a Codual with only the forward data of the tangents, intialised to zero.

DifferentiationInterface/Project.toml
AstitvaAggarwal: yup my bad, if i had added tests this would have come to my attention.

AstitvaAggarwal:
sorry i got preoccupied with some other work, hence the incomplete PR. This would be on route now.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: thanks for the info and pointing this out.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: yeah, ill change it up once its perfect for the single arg case. thanks

gdalle:
Please keep in mind that every commit costs around 6 hours of CI budget. I suggest you make as many modifications as possible locally and add tests first before pushing

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,49 @@
...
+    y = zero_fcodual(f(primal_x))
+
+    # output is a vector, so we need to use the vector pullback
+    function pullback!!(dy::NoRData)
+        tx = DI.pullback(f, backend, primal_x, (fdata(y.dx),))
+        return NoRData(), only(tx)
+    end
+
+    # output is a scalar, so we can use the scalar pullback
+    function pullback!!(dy)
```
yebai: Can we restrict the type for `dy` if it is a scalar?

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -95,7 +95,7 @@ In general, using a forward outer backend over a reverse inner backend will yiel
 The wrapper [`DifferentiateWith`](@ref) allows you to switch between backends.
 It takes a function `f` and specifies that `f` should be differentiated with the substitute backend of your choice, instead of whatever true backend the surrounding code is trying to use.
 In other words, when someone tries to differentiate `dw = DifferentiateWith(f, substitute_backend)` with `true_backend`, then `substitute_backend` steps in and `true_backend` does not dive into the function `f` itself.
-At the moment, `DifferentiateWith` only works when `true_backend` is either [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl) or a [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible backend.
+At the moment, `DifferentiateWith` only works when `true_backend` is either [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl), [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl) or a [Mooncake.jl](https://github.com/compintell/Mooncake.jl)-compatible backend.
```
yebai:
```suggestion
At the moment, `DifferentiateWith` only works when `true_backend` is either [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl), [Mooncake.jl](https://github.com/compintell/Mooncake.jl), or a [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible backend (e.g., [Zygote.jl](https://github.com/FluxML/Zygote.jl)).
```

DifferentiationInterface/docs/src/faq/differentiability.md
```diff
@@ -111,4 +111,4 @@ There are, however, translation utilities:
 ### Backend switch
 
 Also note the existence of [`DifferentiationInterface.DifferentiateWith`](@ref), which allows the user to wrap a function that should be differentiated with a specific backend.
-Right now it only targets ForwardDiff.jl and ChainRulesCore.jl, but PRs are welcome to define Enzyme.jl and Mooncake.jl rules for this object.
\ No newline at end of file
+Right now it only targets ForwardDiff.jl, ChainRulesCore.jl and Mooncake.jl but PRs are welcome to define Enzyme.jl rules for this object.
```
yebai: It's better to use either `ChainRulesCore.jl` or `ChainRules.jl` consistently

```suggestion
Right now, it only targets ForwardDiff.jl, Mooncake.jl, ChainRules.jl-compatible backends (e.g., [Zygote.jl](https://github.com/FluxML/Zygote.jl)),  but PRs are welcome to define Enzyme.jl rules for this object.
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,48 @@
+@is_primitive MinimalCtx Tuple{DI.DifferentiateWith,<:Union{Number,AbstractArray}}
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, x::CoDual{<:Number})
+    primal_func = primal(dw)
+    primal_x = primal(x)
+    (; f, backend) = primal_func
+    y = zero_fcodual(f(primal_x))
+
+    # output is a vector, so we need to use the vector pullback
```
gdalle, line 25: I don't think it's a good idea to define two methods for `pullback!!`. We should probably define `pullback_scalar!!` and `pullback_array!!`, then return one of those and only one.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,48 @@
+@is_primitive MinimalCtx Tuple{DI.DifferentiateWith,<:Union{Number,AbstractArray}}
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, x::CoDual{<:Number})
+    primal_func = primal(dw)
+    primal_x = primal(x)
+    (; f, backend) = primal_func
+    y = zero_fcodual(f(primal_x))
```
gdalle, line 23: I'm not sure about this, to me it seems like we're forgetting to carry around the `FData` of `y`.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,48 @@
+@is_primitive MinimalCtx Tuple{DI.DifferentiateWith,<:Union{Number,AbstractArray}}
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, x::CoDual{<:Number})
+    primal_func = primal(dw)
```
gdalle, line 20: What if there are derivatives inside `dw`?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,48 @@
...
+    primal_x = primal(x)
+    (; f, backend) = primal_func
+    y = zero_fcodual(f(primal_x))
+
+    # output is a vector, so we need to use the vector pullback
+    function pullback!!(dy::NoRData)
+        tx = DI.pullback(f, backend, primal_x, (fdata(y.dx),))
+        return NoRData(), only(tx)
```
gdalle: We're not sure that `f` has no `RData`

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,48 @@
...
+    return y, pullback!!
+end
+
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, x::CoDual{<:AbstractArray})
+    primal_func = primal(dw)
+    primal_x = primal(x)
+    fdata_arg = fdata(x.dx)
+    (; f, backend) = primal_func
+    y = zero_fcodual(f(primal_x))
+    # in case x is mutated in f calls
```
gdalle: This should not be allowed, at least in the supported DI API (`x` is never mutated in DI)

gdalle (changes requested):
Thanks, we're getting closer!
Unfortunately I think my existing tests are not enough to capture everything that can go wrong in a Mooncake rule. Perhaps the Mooncake test utilities should be brought in, or more sophisticated tests should be written.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: im not sure i understand what you mean by this. what ive tried here is instead of creating the output Codual later, we create it early so that in cases of memory address function return types, the pullback functions can have access to the FData to create the correct adjoints.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: im not sure i understand. dw would be a function and if derivatives are used inside i think that would be handled by the substitute backend?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: image
I think all functions would have no RData

gdalle: Not all callable objects are functions though.
```julia
julia> struct Multiplier
           a::Float64
       end

julia> (m::Multiplier)(x) = m.a * x

julia> m = Multiplier(2)
Multiplier(2.0)

julia> m(3)
6.0

julia> m isa Function
false
```

gdalle: And I'm not even convinced the current Mooncake behavior is correct, see https://github.com/compintell/Mooncake.jl/issues/557

AstitvaAggarwal: Nice catch, Mooncake treats all callable objects as a function, giving `NoRData()`. The issue you have opened in mooncake, i think would need to be figured out independent of DI (atleast this PR) as of now (comes in when Mooncake is a substitute backend as well).

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
gdalle: What I mean is that outside of this code, Mooncake may be trying to differentiate something with respect to some parameters that are stored inside `dw`. We could exclude it, and say that functions passed to `DifferentiateWith` must not close over any active data (that is what I implicitly assume for the chain rule). But let's document it then

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
gdalle: See my remark above, let's assume for simplicity that `f` contains no differentiated data, but with a corresponding warning in the docstring of `DifferentiateWith`

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
gdalle: ok I think I get it

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,46 @@
...
+function Mooncake.rrule!!(dw::CoDual{<:DI.DifferentiateWith}, x::CoDual{<:Number})
+    primal_func = primal(dw)
+    primal_x = primal(x)
+    (; f, backend) = primal_func
+    y = zero_fcodual(f(primal_x))
+
+    # output is a vector, so we need to use the vector pullback
+    function pullback_array!!(dy::NoRData)
+        tx = DI.pullback(f, backend, primal_x, (fdata(y.dx),))
+        return NoRData(), only(tx)
```
gdalle: We should probably ensure that `only(tx) isa rdata_type(typeof(x))`, and at least error otherwise. For simple numbers and arrays it holds, but not in general

DifferentiationInterface/test/Back/DifferentiateWith/test.jl
```diff
@@ -1,11 +1,12 @@
 using Pkg
-Pkg.add(["FiniteDiff", "ForwardDiff", "Zygote"])
+Pkg.add(["FiniteDiff", "ForwardDiff", "Zygote", "Mooncake"])
 
 using DifferentiationInterface, DifferentiationInterfaceTest
 import DifferentiationInterfaceTest as DIT
 using FiniteDiff: FiniteDiff
 using ForwardDiff: ForwardDiff
 using Zygote: Zygote
+using Mooncake: Mooncake
```
gdalle, line 10: In this file we need to add tests that are specific to Mooncake. Ideally I should have done that with the other backends too.
Basically, what we test now is the projection of the Mooncake rule you wrote onto the subset of stuff that DI cares about. But we should also check that the rule is correct from the Mooncake perspective. Probably the best tool for that is [`Mooncake.TestUtils.test_rule`](https://chalk-lab.github.io/Mooncake.jl/stable/utilities/debugging_and_mwes/#Mooncake.TestUtils.test_rule-utilities-debugging_and_mwes)?

DifferentiationInterface/docs/src/faq/differentiability.md
```diff
@@ -111,4 +111,4 @@ There are, however, translation utilities:
 ### Backend switch
 
 Also note the existence of [`DifferentiationInterface.DifferentiateWith`](@ref), which allows the user to wrap a function that should be differentiated with a specific backend.
-Right now it only targets ForwardDiff.jl and ChainRulesCore.jl, but PRs are welcome to define Enzyme.jl and Mooncake.jl rules for this object.
\ No newline at end of file
+Right now, it only targets ForwardDiff.jl, Mooncake.jl, ChainRules.jl-compatible backends (e.g., [Zygote.jl](https://github.com/FluxML/Zygote.jl)),  but PRs are welcome to define Enzyme.jl rules for this object.
```
gdalle: Might as well add the links everywhere then

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -95,7 +95,7 @@ In general, using a forward outer backend over a reverse inner backend will yiel
 The wrapper [`DifferentiateWith`](@ref) allows you to switch between backends.
 It takes a function `f` and specifies that `f` should be differentiated with the substitute backend of your choice, instead of whatever true backend the surrounding code is trying to use.
 In other words, when someone tries to differentiate `dw = DifferentiateWith(f, substitute_backend)` with `true_backend`, then `substitute_backend` steps in and `true_backend` does not dive into the function `f` itself.
-At the moment, `DifferentiateWith` only works when `true_backend` is either [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl) or a [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible backend.
+At the moment, `DifferentiateWith` only works when `true_backend` is either [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl), [Mooncake.jl](https://github.com/compintell/Mooncake.jl), or a [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl)-compatible backend (e.g., [Zygote.jl](https://github.com/FluxML/Zygote.jl)).
```
gdalle: Update to `chalk-lab`

gdalle (changes requested):

DifferentiationInterface/test/Back/DifferentiateWith/test.jl
AstitvaAggarwal: yup, will check it out

AstitvaAggarwal: Okay so it fails for some of the general primal, functions. But for this PR its maybe okay? as `DifferentiateWith` is exclusive to DI, so the user is anyways limited to DI when using the Mooncake substitute backend.

yebai: > it fails for some of the general primal, functions.

To clarify, these (primal) functions are permitted by the DI interface, right? 

Assuming that is true, @gdalle, I think this is okay.

yebai: @AstitvaAggarwal is it possible to add these tests (excluding those not supported by DI) to this PR?

AstitvaAggarwal: Many of them are not, and the ones that are-seem to be present in current tests. but as suggested I'll add them here as well, it would make this rrule better anyways

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -1,5 +1,6 @@
 @is_primitive MinimalCtx Tuple{DI.DifferentiateWith,<:Union{Number,AbstractArray,Tuple}}
 
+# nested vectors, similar are not supported
```
yebai: This is not clear. More explanations and examples are better.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -10,31 +11,41 @@ function Mooncake.rrule!!(
...
+        pullback_nodiff!!
+    elseif typeof(primal(y)) <: Number
         pullback_scalar!!
     elseif typeof(primal(y)) <: Array
         pullback_array!!
-    else
+    elseif typeof(primal(y)) <: Tuple
         pullback_tuple!!
+    else
+        error("$(typeof(primal(y))) primal type currently not supported.")
```
yebai: Can you also print the type information of `f` and `x`, which are helpful for debugging.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -43,40 +54,50 @@ end
...
+    elseif typeof(primal(y)) <: Number
         pullback_scalar!!
-    elseif typeof(primal(y)) <: Array
+    elseif typeof(primal(y)) <: AbstractArray
         pullback_array!!
-    else
+    elseif typeof(primal(y)) <: Tuple
         pullback_tuple!!
+    else
+        error("$(typeof(primal(y))) primal type currently not supported.")
```
yebai: Similarly, can you also print the type information of f and x, which are helpful for debugging?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -89,90 +110,93 @@ end
...
-    #         false
-    #     end
-    # end
-    # for test in additional_test_set
-    #     if is_valid(test[2])
-    #         push!(test_cases, test)
-    #     end
-    # end
+
+    map([(x) -> DI.DifferentiateWith(x, DI.AutoZygote())]) do F
```
yebai: Better to avoid `Zygote`.

```suggestion
    map([(x) -> DI.DifferentiateWith(x, DI.ChainRules())]) do F
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -89,90 +110,93 @@ end
...
+                test_cases,
+                Any[
+                    (false, :stability, nothing, F(Base.FastMath.sincos), P(3.0)),
+                    (false, :none, nothing, F(Mooncake.__vec_to_tuple), Any[P(1.0)]),
+                ]...,
+            )
+        end
+    end
+
+    map([(x) -> DI.DifferentiateWith(x, DI.AutoZygote())]) do F
```
yebai: Better to avoid `Zygote`.

```suggestion
    map([(x) -> DI.DifferentiateWith(x, DI.ChainRules())]) do F
```

DifferentiationInterface/test/Back/DifferentiateWith/test.jl
```diff
@@ -31,6 +31,6 @@ test_differentiation(
     logging=LOGGING,
 )
 
-@testset "new" begin
+@testset "Mooncake tests" begin
     Mooncake.TestUtils.run_rrule!!_test_cases(StableRNG, Val(:diffwith))
```
yebai: Can you also check the new rule's type stability, e.g., using JET? If I recall correctly, Mooncake might have the tools for this.

yebai (commented):
Good work, @AstitvaAggarwal. We are getting there. I have just a few more quality-related comments on the documents, type stability, etc.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
AstitvaAggarwal: Yeah my bad, i was gonna add more info about this after the inital reviews over current tests passes. will fix this. lmk if the current comment suffices.

DifferentiationInterface/test/Back/DifferentiateWith/test.jl
AstitvaAggarwal: yes, for each test case we would pass the flags  `:stability`, `:allocs` or `:stability_and_allocs`. Ive updated the flags for all the cases, with almost all (except 1) having atleast type stability checks.

yebai (approved):
Thanks @AstitvaAggarwal for the nice work! 

EDIT: I am happy with this PR, but still need @gdalle's blessing before merging.

DifferentiationInterface/docs/src/faq/differentiability.md
```diff
@@ -111,4 +111,5 @@ There are, however, translation utilities:
 ### Backend switch
 
 Also note the existence of [`DifferentiationInterface.DifferentiateWith`](@ref), which allows the user to wrap a function that should be differentiated with a specific backend.
-Right now it only targets ForwardDiff.jl and ChainRulesCore.jl, but PRs are welcome to define Enzyme.jl and Mooncake.jl rules for this object.
\ No newline at end of file
+
+Right now, it only targets [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl), [Mooncake.jl](), [ChainRules.jl](https://juliadiff.org/ChainRulesCore.jl/stable/)-compatible backends (e.g., [Zygote.jl](https://github.com/FluxML/Zygote.jl)),  but PRs are welcome to define Enzyme.jl rules for this object.
```
gdalle:
```suggestion
Right now, it only targets [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl), [Mooncake.jl](), [ChainRules.jl](https://juliadiff.org/ChainRulesCore.jl/stable/)-compatible backends (e.g., [Zygote.jl](https://github.com/FluxML/Zygote.jl)), but PRs are welcome to define Enzyme.jl rules for this object.
```

DifferentiationInterface/src/misc/differentiate_with.jl
```diff
@@ -13,9 +13,13 @@ Moreover, any larger algorithm `alg` that calls `f2` instead of `f` will also be
...
+    It only makes these functions differentiable if the true backend is either [ForwardDiff](https://github.com/JuliaDiff/ForwardDiff.jl), [Mooncake](https://github.com/chalk-lab/Mooncake.jl) or automatically importing rules from [ChainRules](https://github.com/JuliaDiff/ChainRules.jl) (e.g. [Zygote](https://github.com/FluxML/Zygote.jl)). Some backends are also able to [manually import rules](https://juliadiff.org/ChainRulesCore.jl/stable/#Packages-supporting-importing-rules-from-ChainRules.) from ChainRules.
     For any other true backend, the differentiation behavior is not altered by `DifferentiateWith` (it becomes a transparent wrapper).
 
+!!! warning
+    When using Mooncake as a substitute backend via `DifferentiateWith(f, AutoMooncake())`. The function `f` must not close over any active data.
+    As of now, we cannot differentiate with respect to parameters stored inside `f`.
```
gdalle:
```suggestion
    When using `DifferentiateWith(f, AutoSomething())`, the function `f` must not close over any active data.
    As of now, we cannot differentiate with respect to parameters stored inside `f`.
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
+@is_primitive MinimalCtx Tuple{DI.DifferentiateWith,<:Union{Number,AbstractArray,Tuple}}
+
+# nested vectors (eg. [[1.0]]), Tuples (eg. ((1.0,),)) or similar (eg. [(1.0,)]) primal types are not supported by DI yet !
+# This is because basis construction (DI.basis) does not have overloads for these types.
+# For details, refer commented out test cases to see where the pullback creation fails.
```
gdalle: The issue is that we're testing `DifferentiateWith(f, substitute_backend)` with `substitute_backend = AutoFiniteDiff()`, aka a forward-mode backend. I think it should work with `DifferentiateWith(f, AutoEnzyme())`?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
+@is_primitive MinimalCtx Tuple{DI.DifferentiateWith,<:Union{Number,AbstractArray,Tuple}}
```
gdalle: I find it a bit weird to have this union of `Number, AbstractArray` (two types which are theoretically supported for `x` inputs in DI) and then just `Tuple` (which is not officially part of the supported inputs). Why not also `NamedTuple` for instance? Is it better if we just say `Any`? Or restrict to `Number` and `AbstractArray` for the time being?

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+
+    # inputs are non Differentiable
+    function pullback_nodiff!!(dy::NoRData)
+        @assert tangent_type(typeof(primal(x))) <: NoTangent
+        return NoRData(), dy
+    end
+
+    pullback = if tangent_type(typeof(primal(x))) <: NoTangent
+        pullback_nodiff!!
+    elseif typeof(primal(y)) <: Number
```
gdalle:
```suggestion
    elseif primal(y) isa Number
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_scalar!!
+    elseif typeof(primal(y)) <: Array
```
gdalle:
```suggestion
    elseif primal(y) <: AbstractArray
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_array!!
+    elseif typeof(primal(y)) <: Tuple
```
gdalle:
```suggestion
    elseif primal(y) <: Tuple
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_tuple!!
+    else
+        error(
+            "For the function type $(typeof(primal_func)) and input type $(typeof(primal_x)), the primal type $(typeof(primal(y))) is currently not supported.",
+        )
```
gdalle: We should throw a specific error type, not just the default `error`. Something `MooncakeDifferentiateWithError`

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        @assert rdata(only(tx)) isa rdata_type(tangent_type(typeof(primal_x)))
+        return NoRData(), rdata(only(tx))
+    end
+
+    # output is a Tuple, NTuple
+    function pullback_tuple!!(dy::Tuple)
+        tx = DI.pullback(f, backend, primal_x, (dy,))
+        @assert rdata(only(tx)) isa rdata_type(tangent_type(typeof(primal_x)))
+        return NoRData(), rdata(only(tx))
+    end
```
gdalle: This only works for tuples of numbers, right? With a tuple of arrays for instance, it would fail? Perhaps it would be best for us to just remove support for tuples completely at first.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
gdalle:
```julia
struct MooncakeDifferentiateWithError <: Exception
    msg::String
end

Base.showerror(io::IO, e::MooncakeDifferentiateWithError) = print(io, "MooncakeDifferentiateWithError: ", e.msg)
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+
+    # inputs are non Differentiable
+    function pullback_nodiff!!(dy::NoRData)
+        @assert tangent_type(typeof(primal(x))) <: Vector{NoTangent}
+        return NoRData(), dy
+    end
+
+    pullback = if tangent_type(typeof(primal(x))) <: Vector{NoTangent}
+        pullback_nodiff!!
+    elseif typeof(primal(y)) <: Number
```
gdalle:
```suggestion
    elseif primal(y) isa Number
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_scalar!!
+    elseif typeof(primal(y)) <: AbstractArray
```
gdalle:
```suggestion
    elseif primal(y) <: AbstractArray
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_array!!
+    elseif typeof(primal(y)) <: Tuple
```
gdalle:
```suggestion
    elseif primal(y) <: Tuple
```

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_tuple!!
+    else
+        error(
+            "For the function type $(typeof(primal_func)) and input type $(typeof(primal_x)), the primal type $(typeof(primal(y))) is currently not supported.",
+        )
```
gdalle: Same remark about custom error

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        return NoRData(), NoRData()
+    end
+
+    # output is a Tuple, NTuple
+    function pullback_tuple!!(dy::Tuple)
+        tx = DI.pullback(f, backend, primal_x, (dy,))
+        @assert rdata(first(only(tx))) isa rdata_type(tangent_type(typeof(first(primal_x))))
+        fdata_arg .+= only(tx)
+        return NoRData(), NoRData()
+    end
```
gdalle: Same remark about tuples

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+        pullback_tuple!!
+    else
+        error(
+            "For the function type $(typeof(primal_func)) and input type $(typeof(primal_x)), the primal type $(typeof(primal(y))) is currently not supported.",
+        )
+    end
+
+    return y, pullback
+end
+
```
gdalle: Starting here, the testing code should go in the testing file

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,268 @@
...
+function Mooncake.generate_derived_rrule!!_test_cases(rng_ctor, ::Val{:diffwith})
+    return Any[], Any[]
+end
```
gdalle: Starting here, the testing code relies a lot on very deep Mooncake internals, which are likely to break or get renamed without notice (like the name of a particular test scenario). I was thinking we could just use `Mooncake.TestUtils.test_rule` from `DifferentiationInterfaceTest.Scenario` objects. Will try to cook up an MWE.

gdalle (changes requested):
Very good work @AstitvaAggarwal, thank you so much for taking the time! My main remaining questions are about:

- support for tuples: do we keep it, do we remove it? It seems a bit arbitrary to have just tuples in addition to the "officially" supported `Number` and `AbstractArray`, especially if these can only be tuples of numbers
- testing code, which should not touch on so many Mooncake internals

To lighten the load on you, I'll make the suggested modifications myself, and you can tell me what you think after! Thanks again for your patience.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
gdalle: Resolved by removing tuples for the time being

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/differentiate_with.jl
```diff
@@ -0,0 +1,85 @@
...
+
+function Base.showerror(io::IO, e::MooncakeDifferentiateWithError)
+    return print(
+        io,
+        "MooncakeDifferentiateWithError: For the function type $(e.F) and input type $(e.X), the output type $(e.Y) is currently not supported.",
+    )
+end
+
+# For details, refer commented out test cases to see where the pullback creation fails.
```
gdalle:
```suggestion
```

gdalle (approved):

PR #772, opened 2025-04-10
perf: remove Mooncake warmup during preparation

gdalle:
Mooncake preparation used to run the pullback or gradient once in order to compile the rule. This led to a x2 slowdown when used without preparation, because
```julia
DI.gradient(f, backend, x)
```
falls back on
```julia
prep = DI.prepare_gradient(f, backend, x)  # used to include one gradient call
DI.gradient(f, prep, x)
```
This PR removes the duplicate, so the first run of `DI.gradient` will be slower but only due to compilation.

Ping @penelopeysm who had asked about this


willtebbutt (approved):
LGTM -- thanks for looking into this.

PR #773, opened 2025-04-10
Fix typo in Scenario docstring

penelopeysm:
Noticed this while looking through the Scenario source code (`args` having been removed in #470).


gdalle (approved):
thanks!

PR #788, opened 2025-05-09
feat: error hints for Enzyme

gdalle:
@vchuravy @wsmoses are there other Enzyme errors for which we can provide DI-specific hints to users?


DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/init.jl
```diff
@@ -0,0 +1,44 @@
...
+    # see https://github.com/JuliaLang/julia/issues/58367 for why this isn't easier
+    for n in names(Enzyme.Compiler; all=true)
+        T = getfield(Enzyme.Compiler, n)
+        if T isa DataType && T <: Enzyme.Compiler.EnzymeError
+            # robust against internal changes
+            Base.Experimental.register_error_hint(T) do io, exc
+                if occursin("EnzymeMutabilityException", string(nameof(T)))
+                    printstyled(
+                        io,
+                        "\nIf you are using Enzyme through DifferentiationInterface, you may want to try modifying the ADTypes backend object as follows:";
```
wsmoses: This is independent of DI (eg using optimization.jl), can we reword to saying if using an ADType change to?

gdalle: Given that DI pretty much defines the semantics of ADTypes I feel like that is potato-potahto?
Furthermore, in the case of Optimization, I don't think the Enzyme extension actually takes the `function_annotation` setting into account, since it bypasses DI entirely. So I would say the current formulation is accurate.

gdalle: Also, it seems that same extension to Optimization systematically activates runtime activity in the Enzyme mode whether you ask for it or not?
In a nutshell, these hints only make sense when people use the ADTypes backends through DI, so that's why I'd rather mention DI in the message too.

wsmoses: If ADTypes is defined by an external package that feels like a bigger problem.

In any case, given that this error message will be unconditionally printed whenever DI is loaded (even if not used) I'd prefer to be a bit more conservative?

Though that makes me wonder if there's a way we can conditionally check depending on the current error backtrace and change error then

gdalle: > If ADTypes is defined by an external package that feels like a bigger problem.

I agree, that's why in the long run I want to bring ADTypes into the DI repo. Historically, ADTypes pre-existed DI, so when I started DI the goal was to become compatible with the existing semantics (used in SciML, Turing and elsewhere). Now it's pretty much the other way around, with new features of ADTypes that exist only through their DI translation.

This makes DI pretty much the reference implementation of ADTypes, if not strictly speaking the only one. Code bypassing it is of course possible, but it may not be as thorough or well-tested: for instance, OptimizationBase does not fully respect the `mode` kwarg of `AutoEnzyme` (adding runtime activity by default), and completely disregards the `function_annotation` kwarg.

> In any case, given that this error message will be unconditionally printed whenever DI is loaded (even if not used) I'd prefer to be a bit more conservative?

I made the message more explicit.

gdalle: > Though that makes me wonder if there's a way we can conditionally check depending on the current error backtrace and change error then

That's a lot more trouble than it's worth, I think the current behavior is good enough.

Issue #800, opened 2025-05-16
is preparation thread safe?


FerreolS:
Hi,
It is more a question than a issue as I didn't see any information in the doc:
Is the preparation thread safe? 
This kind of code seems to work but is it legit for all possible backend? 

```julia-repl
julia> backendZ = AutoZygote()
AutoZygote()

julia> backendFD = AutoForwardDiff()
AutoForwardDiff()

julia> prepFD = prepare_gradient(sin, backendFD, x)
DifferentiationInterface.PullbackGradientPrep{Nothing, Float64, DifferentiationInterface.PushforwardPullbackPrep{Nothing, DifferentiationInterfaceForwardDiffExt.ForwardDiffOneArgPushforwardPrep{Nothing, ForwardDiff.Tag{typeof(sin), Float64}, ForwardDiff.Dual{ForwardDiff.Tag{typeof(sin), Float64}, Float64, 1}, Tuple{}}}}(Val{Nothing}(), 0.8414709848078965, DifferentiationInterface.PushforwardPullbackPrep{Nothing, DifferentiationInterfaceForwardDiffExt.ForwardDiffOneArgPushforwardPrep{Nothing, ForwardDiff.Tag{typeof(sin), Float64}, ForwardDiff.Dual{ForwardDiff.Tag{typeof(sin), Float64}, Float64, 1}, Tuple{}}}(Val{Nothing}(), DifferentiationInterfaceForwardDiffExt.ForwardDiffOneArgPushforwardPrep{Nothing, ForwardDiff.Tag{typeof(sin), Float64}, ForwardDiff.Dual{ForwardDiff.Tag{typeof(sin), Float64}, Float64, 1}, Tuple{}}(Val{Nothing}(), ForwardDiff.Tag{typeof(sin), Float64}, Dual{ForwardDiff.Tag{typeof(sin), Float64}}(1.0,1.0), ())))

julia> prepZ = prepare_gradient(sin, backendZ, x)
DifferentiationInterface.NoGradientPrep{Nothing}(Val{Nothing}())

julia> a = randn(1000);

julia> grad = similar(a);

julia> Threads.@threads for i in eachindex(a,grad); grad[i] = DifferentiationInterface.gradient(sin, prepZ, backendZ,a[i]) ; end

julia> Threads.@threads for i in eachindex(a,grad); grad[i] = DifferentiationInterface.gradient(sin, prepFD, backendFD,a[i]) ; end
```

[...]

PR #801, opened 2025-05-16
docs: warn that preparation is not thread-safe

FerreolS:
Here are a few lines about thread safety of preparation for the documentation after my question #800

### DIFF
--- a/DifferentiationInterface/docs/src/explanation/operators.md
+++ b/DifferentiationInterface/docs/src/explanation/operators.md
@@ -117,6 +117,7 @@ op(f, prep, backend, x, [t])  # fast because it skips preparation
 
 !!! warning
     The `prep` object is the last argument before `backend` and it is always mutated, regardless of the bang `!` in the operator name.
+    As a consequence preparation is **not thread safe** and  sharing  `prep` object between threads may lead to undefined behavior.   If you need to run differentiation concurrently, prepare separate `prep` objects for each thread.
 
 ### Reusing preparation
 
--- a/DifferentiationInterface/docs/src/explanation/operators.md
+++ b/DifferentiationInterface/docs/src/explanation/operators.md
@@ -117,7 +117,7 @@ op(f, prep, backend, x, [t])  # fast because it skips preparation
 
 !!! warning
     The `prep` object is the last argument before `backend` and it is always mutated, regardless of the bang `!` in the operator name.
-    As a consequence preparation is **not thread safe** and  sharing  `prep` object between threads may lead to undefined behavior.   If you need to run differentiation concurrently, prepare separate `prep` objects for each thread.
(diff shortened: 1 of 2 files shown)


DifferentiationInterface/src/docstrings.jl
```diff
@@ -25,6 +25,8 @@ function docstring_prepare(operator; samepoint=false, inplace=false)
         The preparation result `prep` is only reusable as long as the arguments to `$operator` do not change type or size, and the function and backend themselves are not modified.
         Otherwise, preparation becomes invalid and you need to run it again.
         In some settings, invalid preparations may still give correct results (e.g. for backends that require no preparation), but this is not a semantic guarantee and should not be relied upon.
+    !!! danger
+        The preparation result `prep` is **not thread safe**. Sharing it between threads may lead to undefined behavior. If you need to run differentiations concurrently, prepare separate `prep` objects for each thread.
```
gdalle:
```suggestion

    !!! danger
        The preparation result `prep` is **not thread-safe**. Sharing it between threads may lead to unexpected behavior. If you need to run differentiations concurrently, prepare separate `prep` objects for each thread.
```

DifferentiationInterface/docs/src/explanation/operators.md
```diff
@@ -117,6 +117,7 @@ op(f, prep, backend, x, [t])  # fast because it skips preparation
 
 !!! warning
     The `prep` object is the last argument before `backend` and it is always mutated, regardless of the bang `!` in the operator name.
+    As a consequence preparation is **not thread safe** and  sharing  `prep` object between threads may lead to undefined behavior.   If you need to run differentiation concurrently, prepare separate `prep` objects for each thread.
```
gdalle:
```suggestion
    As a consequence, preparation is **not thread-safe** and  sharing `prep` objects between threads may lead to unexpected behavior. If you need to run differentiation concurrently, prepare separate `prep` objects for each thread.
```

DifferentiationInterface/src/docstrings.jl
```diff
@@ -26,6 +26,9 @@ function docstring_prepare(operator; samepoint=false, inplace=false)
         Otherwise, preparation becomes invalid and you need to run it again.
         In some settings, invalid preparations may still give correct results (e.g. for backends that require no preparation), but this is not a semantic guarantee and should not be relied upon.
 
+    !!! danger
+        The preparation result `prep` is **not thread-safe**. Sharing it between threads may lead to unexpected behavior. If you need to run differentiations concurrently, prepare separate `prep` objects for each thread.
```
gdalle:
```suggestion
        The preparation result `prep` is **not thread-safe**. Sharing it between threads may lead to unexpected behavior. If you need to run differentiation concurrently, prepare separate `prep` objects for each thread.
```

gdalle (approved):
Thanks, this is a useful addition!

PR #807, opened 2025-05-28
Revert "feat: support DuplicatedNoNeed as function annotation for Enzyme"

gdalle:
Reverts JuliaDiff/DifferentiationInterface.jl#805

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.7.1"
+version = "0.7.0"
 
 [deps]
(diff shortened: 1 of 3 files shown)


wsmoses:
why was this reverted?

gdalle:
Because the tests passed but I don't understand why. Despite the follow-up exchange on Discourse I still don't fully get what I'm allowed to do with `DuplicatedNoNeed`, nor can I think of a meaningful test case where the closure storage is write-only (what good does a write-only storage do?).
The test case I had in the DI test suite writes to closure storage _and then reads back from it_, but switching to `function_annotation=DuplicatedNoNeed` did not make it fail. I'm not sure why, so as a precaution I'm putting correctness over performance until I figure it out.

gdalle:
Essentially I don't want to release something where tests are supposed to fail but succeed and I don't know why.

gdalle:
The test case in question is here:

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/main/DifferentiationInterfaceTest/src/scenarios/modify.jl#L108-L164

I take a standard test function, and then I force both the input and the output to go through a buffer

PR #808, opened 2025-06-02
docs: make Enzyme warning more visible

gdalle:
- Turn the Enzyme warning on the README into a GitHub admonition for more visibility:
- Translate it to a Documenter.jl admonition when building the docs.
- Add a related warning to the backends page

The warning now looks like this on the README:

> [!CAUTION]
> Note that in some cases, going through DifferentiationInterface.jl might be slower or cause more errors than a direct call to the backend’s API. This is especially true for Enzyme.jl, whose handling of activities and multiple arguments is not fully supported here. We are working on this challenge, and welcome any suggestions or contributions. Meanwhile, if differentiation fails or takes too long, consider using Enzyme.jl through its [native API](https://enzymead.github.io/Enzyme.jl/stable/) instead.

Related:

- https://github.com/SciML/ADTypes.jl/pull/109
- https://github.com/EnzymeAD/Enzyme.jl/pull/2415


gdalle:
cc @wsmoses, trying to make this as visible and explicit as I can

Issue #802, opened 2025-05-22
Inconsistency in handling empty arguments


penelopeysm:
When differentiating with respect to an empty array, the results tend to vary:

```julia
using DifferentiationInterface, ForwardDiff, ReverseDiff, Mooncake, Enzyme

ADTYPES = [
    AutoForwardDiff(),
    AutoReverseDiff(),
    AutoMooncake(; config=nothing),
    AutoEnzyme(; mode=Forward),
    AutoEnzyme(; mode=Reverse),
    # and more...
]

for adtype in ADTYPES
    DifferentiationInterface.value_and_gradient(sum, adtype, Float64[])
end
```

ReverseDiff, Mooncake, and reverse Enzyme all happily return `(0.0, [])` 😄 

Forward Enzyme tries to use a batch size of 0 and errors:

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/6a58124902a9cc3e623014db28eacd4b56ba4434/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl#L11-L14

And ForwardDiff tries to construct a `GradientResult` which errors:

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/6a58124902a9cc3e623014db28eacd4b56ba4434/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/onearg.jl#L315-L318

https://github.com/JuliaDiff/DiffResults.jl/blob/fcf7858d393f0597fc74e195ed46f7bcbe5ff66c/src/DiffResults.jl#L64-L65

Funnily enough `gradient` with ForwardDiff (rather than `value_and_gradient`) is fine because it doesn't try to construct the `GradientResult`. I imagine the other operators would also have varying behaviour.

I suppose it is a bit of a trivial edge case, but would it be possible to unify the behaviour of the AD backends?

gdalle:
Hi, thanks for raising this issue! Honestly at this point it doesn't rank very high on my list of priorities, so it may linger for a while.
Do you have any example of realistic scenario where the user would want to compute an empty gradient and rely on it not erroring?

penelopeysm:
> scenario where the user

Mostly, dealing with the case where somebody tries to sample from a model that only contains likelihood terms:

```julia
using Turing

@model function empty(x)
    x ~ Normal()
end

sample(empty(1.0), NUTS(), 100)
```

[...]

PR #835, opened 2025-08-15
fix: improve support for empty inputs (still not guaranteed)

penelopeysm:
I ran into #802 again while refactoring some tests on Bijectors.jl, so this is a patch that lets forward-mode Enzyme work with empty inputs.


gdalle:
I gave it a try too and it's a bit harder than it looks to make this work consistently. First, there are unavoidable discrepancies across backends:

- ForwardDiff can deal with zero chunk sizes, in fact it chooses them voluntarily with `ForwardDiff.pickchunksize(0)`.
- Enzyme cannot deal with zero chunk sizes, it will error [here](https://github.com/EnzymeAD/Enzyme.jl/blob/328927ed5b43a326aff3544e04eb8ba43bb9c073/src/sugar.jl#L408).
- Other backends must have a default chunk size of 1, we cannot use `max(1, length(x))` or it becomes type-unstable (see [here](https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/08d57084d58bda79b1d37a3b3d6f6a4662e5c2a3/DifferentiationInterface/src/utils/batchsize.jl#L49-L55)).

This suggests we need to handle batch sizes of 0 while also setting a minimum of 1 when we can.
Second, there are some difficulties related to the DI operators themselves:

- For wrong-mode pushforwards and pullbacks, we propagate basis vectors, which error when empty (e.g. [here](https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/08d57084d58bda79b1d37a3b3d6f6a4662e5c2a3/DifferentiationInterface/src/first_order/pullback.jl#L288)). This is solvable.
- For Jacobians and Hessians, we precompute a set of basis vectors and perform preparation on the first one, which may not exist when the vector is empty (e.g. [here](https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/08d57084d58bda79b1d37a3b3d6f6a4662e5c2a3/DifferentiationInterface/src/first_order/jacobian.jl#L214-L216)). This is solvable.
- For Jacobians and Hessians, `mapreduce(hcat, collection)` errors when the collection is empty (e.g. [here](https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/08d57084d58bda79b1d37a3b3d6f6a4662e5c2a3/DifferentiationInterface/src/first_order/jacobian.jl#L373-L389)). I'm not sure how to solve that one.

gdalle:
To give you an example of the kind of changes necessary, you can take a look at [this commit](https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/835/commits/8f93335fa941dda2ad4bd68aab1ecb50edfcf9e6). I'm not sure supporting empty arrays is worth it, especially if support is always gonna be divergent across backends as you noted in #802

penelopeysm:
Thanks for looking into it in more detail! I definitely see the issues with making it consistent. I didn't really set out to solve it for all backends, that was a task that was too annoying and possibly also too unimportant -- hence my very tiny single commit 😄. I don't think that DI needs to explicitly guarantee or document that empty inputs will work for all backends; but it's just nice to not get an error and if some code could be tweaked to make that happen, that's a net benefit IMO.

Did that commit alone break ForwardDiff or some other backend? I was sort of hoping that CI would catch issues that I hadn't thought of or didn't know of (of which I'm sure there are plenty). But also my impression was that the lines I changed should only affect empty inputs.

gdalle:
> Did that commit alone break ForwardDiff or some other backend?

I don't think so but I wanted to add proper tests covering this line, and to make sure that the global approach is at least vaguely coherent. I tried defining a batch size of 0 for empty arrays by default, but I think you're right that 1 is better. I ended up adding a bunch of empty test scenarios in DIT to make my life easier. DI's default implementation of `jacobian` still fails on empty vectors because of the uninitialized `mapreduce`, but apart from that this is more satisfactory to me. Can you check whether the added tests cover everything that was failing for you?

gdalle:
Enzyme tests failing due to https://github.com/EnzymeAD/Enzyme.jl/issues/2523

DIT tests failing because I must have introduced a type instability somewhere

PR #839, opened 2025-08-16
feat: compute `Scenario` results with a reference backend

### LINKED ISSUE #834: allow generating `DIT.Scenario` reference results with a given backend
In a lot of our AD testing stuff we want to compare the results against some backend (ideally FiniteDifferences, in practice often ForwardDiff). I wonder if it might be a nice ergonomic improvement to be able to specify that directly somehow when constructing a scenario? i.e. make a one-liner for what's currently

```julia
ref_grad = DifferentiationInterface.gradient(f, reference_adtype, x)
scen = Scenario{:gradient,:out}(f, x; res1=ref_grad)
```

It's certainly easy for me to define a wrapper function but I'm lazy :-)

Although now I look at it, it seems like that might entail some awkward design choices for in-place operations like deepcopying `x`.

gdalle:
Fixes #834

@penelopeysm what do you think?

### DIFF
--- a/DifferentiationInterface/CHANGELOG.md
+++ b/DifferentiationInterface/CHANGELOG.md
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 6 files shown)


penelopeysm:
Sorry I only got to this! I would be super pleased with this :)

gdalle:
I thought you weren't using DITest in Turing?

penelopeysm:
Yup, no DIT in Turing.jl itself (well, not as of the time of writing at least) — but it was [this function in particular](https://github.com/TuringLang/Bijectors.jl/blob/main/test/ad/utils.jl) which made me think that this feature in DIT would be nice :)

PR #841, opened 2025-08-16
chore: test on Julia 1.12 with JET 0.10

gdalle:
Type inference still fails because of https://github.com/aviatesk/JET.jl/issues/699

Test sets are not supported, see https://github.com/aviatesk/JET.jl/issues/742


.github/workflows/Test.yml
```diff
@@ -97,21 +108,26 @@ jobs:
       actions: write
       contents: read
     strategy:
-      fail-fast: true
+      fail-fast: false
```
gdalle:
```suggestion
      fail-fast: true
```

.github/workflows/Test.yml
```diff
@@ -28,8 +28,9 @@ jobs:
       fail-fast: false  # TODO: toggle
```
gdalle:
```suggestion
      fail-fast: true  # TODO: toggle
```

Issue #855, opened 2025-09-29
Failing Enzyme tests


gdalle:
An issue to list some failing Enzyme tests which I deactivated.

Common setup:

```julia
using DifferentiationInterface
import DifferentiationInterfaceTest as DIT
using SparseConnectivityTracer
using SparseMatrixColorings
import Enzyme

function MyAutoSparse(backend)
    return AutoSparse(
        backend;
        sparsity_detector=TracerSparsityDetector(),
        coloring_algorithm=GreedyColoringAlgorithm(),
    )
end
```

Forward-over-forward HVP (see also https://github.com/EnzymeAD/Enzyme.jl/issues/2437):

<details>

```julia
julia> hvp(
           sum,
           SecondOrder(AutoEnzyme(; mode=Enzyme.Forward), AutoEnzyme(; mode=Enzyme.Forward)),
           [1.0],
           ([1.0],),
       )
ERROR: 
No forward mode derivative found for jl_get_builtin_fptr
 at context:   %10 = call {} addrspace(10)* ({} addrspace(10)*, {} addrspace(10)**, i32)* @jl_get_builtin_fptr({} addrspace(10)* noundef addrspacecast ({}* inttoptr (i64 4725148592 to {}*) to {} addrspace(10)*)) #14, !dbg !38

Stacktrace:
 [1] make_context_shadows
   @ ~/Documents/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl:99
 [2] prepare_gradient_nokwarg
   @ ~/Documents/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/forward_onearg.jl:151


Stacktrace:
  [1] make_context_shadows
    @ ~/Documents/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl:99 [inlined]
  [2] prepare_gradient_nokwarg
    @ ~/Documents/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/forward_onearg.jl:151
  [3] gradient
    @ ~/Documents/GitHub/Julia/DifferentiationInterface.jl/DifferentiationInterface/src/first_order/gradient.jl:62
  [4] shuffled_gradient
[...]

PR #856, opened 2025-09-29
test: untoggle newly-passing Enzyme tests

gdalle:
Related:
- #855

### DIFF
--- a/DifferentiationInterface/test/Back/Enzyme/test.jl
+++ b/DifferentiationInterface/test/Back/Enzyme/test.jl
@@ -1,3 +1,5 @@
+# see https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/855
+
 using Pkg
 Pkg.add("Enzyme")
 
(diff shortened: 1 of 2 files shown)


DifferentiationInterface/test/Back/Enzyme/test.jl
```diff
@@ -75,23 +77,19 @@ end;
...
 @testset "Second order" begin
+    forward_enzyme = AutoEnzyme(; mode=Enzyme.Reverse)
+    reverse_enzyme = AutoEnzyme(; mode=Enzyme.Reverse)
     test_differentiation(
         [
             AutoEnzyme(),
-            SecondOrder(
-                AutoEnzyme(; mode=Enzyme.Reverse), AutoEnzyme(; mode=Enzyme.Forward)
-            ),
+            # SecondOrder(forward_enzyme, reverse_enzyme),  # TODO: toggle
```
wsmoses: can these be marked as test_broken instead of just not run?

gdalle: No because DifferentiationInterfaceTest doesn't currently offer the option to set some of the tests as broken

PR #877, opened 2025-10-29
chore: update JET & JLArrays compat

gdalle:
- Bump JET to 0.11 (only on Julia 1.12)
- Bump JLArrays to 0.3
- Include test and docs deps in CompatHelper

Supersedes #872 and #862


gdalle:
Failing because of https://github.com/JuliaLang/Pkg.jl/issues/4442

PR #908, opened 2025-11-05
Use `FiniteDiff.finite_difference_jacobian!`

devmotion:
Similar to the existing in-place methods that use `finite_difference_gradient!` and `finite_difference_hessian!`.


devmotion:
Was looking at the wrong file.

PR #914, opened 2025-11-12
test: remove neural network tests, clean up envs

gdalle:
- Remove tests on Lux and Flux scenarios since neither is using or recommending DI to compute gradients
- Clean up DI's test env so that DIT is natively inside it


gdalle:
Running into https://github.com/JuliaLang/Pkg.jl/issues/1585 again

PR #325, opened 2024-06-20
Reactant prototype

gdalle:
First trial for #265

@wsmoses I run into a weird world age error in the test, any clue?

Issue #265, opened 2024-05-17
Add Reactant


wsmoses:
It's not quite an AD engine in its own right, but given how it traces things and can make for much better performance/compatibility, and can be used with an AD engine, it seems like a natural fit.

gdalle:
Why not but I don't understand what the package does or how it would fit within DI, so you'll have to hold my hand on this one ^^

wsmoses:
Yeah so basically, Reactant exports a compile utility which takes a julia function and arguments, and compiles it to a nice/fast version of it. It does require that all data one cares about to be passed in using ConcreteRArrays rather than Ararys [and structs of those/etc]. We do have a tracer utility which will take an arg and auotmatically changes arrays to ConcreteRArrays for you though.

This can obviously compile code with an autodiff call on the inside, such as Enzyme. See here: https://github.com/EnzymeAD/Reactant.jl/blob/292dc03593ceb1a7a1f022fd7d3289bd69b000b5/test/basic.jl#L81. Enzyme usage should end up with quite good performance since the fancy optimizations that Reactant applies will be able to interoperate with AD.

wsmoses:
We also now have a brief tldr in the readme: https://github.com/EnzymeAD/Reactant.jl/tree/main

gdalle:
So how would you see this in the context of DI? As a variant of the Enzyme backend which Reactant-compiles the function during preparation?

wsmoses:
I think the other way around. Reactant can take another AD backend (perhaps
with enzyme as default)

gdalle:
And do what with it? Compile and convert everything before and after use?

wsmoses:
Yup

gdalle:
I would welcome a PR with an example!

wsmoses:
I'm not quite sure how to make an example PR, but here's an example from our tests: https://github.com/EnzymeAD/Reactant.jl/blob/59d9304948cc7e28acdd4351db5e069d62a4f4ec/test/basic.jl#L92

```julia

function sumcos(x)
    return sum(cos.(x))
end

function resgrad_ip(x)
    dx = Enzyme.make_zero(x)
    res = Enzyme.autodiff(ReverseWithPrimal, sumcos, Active, Duplicated(x, dx))
    return (res, dx)
end

 c = Reactant.ConcreteRArray(ones(3, 2))
    f = Reactant.compile(resgrad_ip, (c,))
    orig, r = f(c)

[...]

PR #918, opened 2025-11-15
feat: add gradient with AutoReactant

gdalle:
Updated experiments with Reactant-accelerated derivatives.

@wsmoses is this still the right paradigm in your opinion? I may not implement every operator right away but I thought starting with a gradient made sense

Related:

- https://github.com/SciML/ADTypes.jl/pull/127
- #265
- #325

>[!WARNING]
> Re-toggle tests once this is mergeable


DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,81 @@
...
+function DI.prepare_gradient_nokwarg(strict::Val, f::F, rebackend::AutoReactant, x) where {F}
+    _sig = DI.signature(f, rebackend, x; strict)
+    backend = rebackend.mode
+    xr = to_rarray(x)
+    gr = to_rarray(similar(x))
+    _gradient(_xr) = DI.gradient(f, backend, _xr)
+    _gradient!(_gr, _xr) = copy!(_gr, DI.gradient(f, backend, _xr))
+    _value_and_gradient(_xr) = DI.value_and_gradient(f, backend, _xr)
+    function _value_and_gradient!(_gr, _xr)
+        y, __gr = DI.value_and_gradient(f, backend, _xr)
```
wsmoses: you should avoid the use of closures here it will cause problems

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,81 @@
...
+    compiled_gradient::CG
+    compiled_gradient!::CG!
+    compiled_value_and_gradient::CVG
+    compiled_value_and_gradient!::CVG!
+end
+
+function DI.prepare_gradient_nokwarg(strict::Val, f::F, rebackend::AutoReactant, x) where {F}
+    _sig = DI.signature(f, rebackend, x; strict)
+    backend = rebackend.mode
+    xr = to_rarray(x)
```
wsmoses: you shoudl check if x / gr are already reactant arrays, if so we should not call to_rarray and/or copy

gdalle: I was wondering whether the copy back from RArray was necessary at all in the general case. Are there limitations on what Reactant arrays can do?

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: What kind of problems? This works in my small test

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
wsmoses: so Reactant Arrays are data that REactant owns, it can be on device (like gpu), distributed across devices/etc. It is also notably what reactant tracks.

See https://enzymead.github.io/Reactant.jl/dev/tutorials/partial-evaluation

The final result doesn't need to be copied back, but the input should be changed to a reactant array if not already (and thus if the op expected in place for the input then it needs to be copied back). In contrast if returning a new array, returning the reactant one is fine

gdalle: Doesn't `to_rarray` compile to a no-op if the input already is an RArray?

wsmoses: That's not something we should assume here

gdalle: But if that's not yet the case, can't we add that one line to Reactant? Is there an issue here that I don't see?

gdalle: ’’’
to_rarray(x::RArray) = x
’’’

wsmoses: The semantics of to_rarray currently does not guarantee if the object returned will be the same or a copy, and we shouldn't rely on it since Reactant may choose to specify this differently in a non breaking change

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: I am now avoiding closures by directly compiling the DI functions, and possibly partially evaluating `f` and `backend` (by not converting them to `RArray`)

gdalle:
@wsmoses does this look better to you now?
I'm not sure what we should do in terms of storage versus allocations. We can store `xr` (and even `contextsr`) during preparation and then copy to them at execution time instead of generating a new `RArray`, but that would require a copying method to be defined (which doesn't apply to all non-array objects).

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,80 @@
...
+    )
+end
+
+function DI.gradient(
+        f::F, prep::ReactantGradientPrep, rebackend::AutoReactant, x, contexts::Vararg{DI.Context, C}
+    ) where {F, C}
+    DI.check_prep(f, prep, rebackend, x)
+    backend = rebackend.mode
+    (; xr, compiled_gradient) = prep
+    copyto!(xr, x)
```
wsmoses: We should only do this if x is not a reactantarray

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,80 @@
...
+    compiled_value_and_gradient::CVG
+    compiled_value_and_gradient!::CVG!
+end
+
+function DI.prepare_gradient_nokwarg(
+        strict::Val, f::F, rebackend::AutoReactant, x, contexts::Vararg{DI.Context, C}
+    ) where {F, C}
+    _sig = DI.signature(f, rebackend, x; strict)
+    backend = rebackend.mode
+    xr = to_reac(x)
```
wsmoses: We shouldn't save anything as a prep argument if a reactant array, I would keep this as if reactant array then xr is nothing otherwise to_rarray(x)

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,80 @@
...
+    return gr
+end
+
+function DI.value_and_gradient(
+        f::F, prep::ReactantGradientPrep, rebackend::AutoReactant, x, contexts::Vararg{DI.Context, C}
+    ) where {F, C}
+    DI.check_prep(f, prep, rebackend, x)
+    backend = rebackend.mode
+    (; xr, compiled_value_and_gradient) = prep
+    copyto!(xr, x)
```
wsmoses: Same comment here

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,80 @@
...
+    return yr, gr
+end
+
+function DI.gradient!(
+        f::F, grad, prep::ReactantGradientPrep, rebackend::AutoReactant, x, contexts::Vararg{DI.Context, C}
+    ) where {F, C}
+    DI.check_prep(f, prep, rebackend, x)
+    backend = rebackend.mode
+    (; xr, gr, compiled_gradient!) = prep
+    copyto!(xr, x)
```
wsmoses: Etc

DifferentiationInterface/test/Back/Reactant/test.jl
```diff
@@ -0,0 +1,21 @@
...
+using Reactant
+using Test
+
+backend = AutoReactant()
+
+@test check_available(backend)
+@test check_inplace(backend)
+
+test_differentiation(
+    backend, DifferentiationInterfaceTest.default_scenarios(;
```
wsmoses, line 16: Can you add a test that the prep contains no data except the compiled fn if compiled for a reactant array

wsmoses:
Modulo some comments being addressed above this looks reasonable to me.

However note that there may be a potential mismatch in expectations behind what prepare gradient defines and what reactant compile defines.

See https://enzymead.github.io/Reactant.jl/dev/tutorials/partial-evaluation

Currently any data inside a constant or cache will be baked into the compiled function and will not be re read in later evaluation.

Enzyme.jl in particular does not have any such constraint (as it will always re run with live data as prep is nothing).

Something like reversediff compiled probably does bake in the assumption from compilation.

So this is a question of what is the semantics of prep.

If the non differentiated data is the same between prep and evaluation there is no difference between the two

gdalle:
> So this is a question of what is the semantics of prep. If the non differentiated data is the same between prep and evaluation there is no difference between the two

The semantics of `prep`: differentiated and non-differentiated data are free to change between preparation and execution, as long as they keep the same types and sizes. See [here](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/explanation/operators/#Reusing-preparation) for details.

> Currently any data inside a constant or cache will be baked into the compiled function and will not be re read in later evaluation.

I thought converting the `contexts` into reactant arrays inside `contextr` would allow them to be traced? If that's true, then they won't be partially evaluated into the compiled function, which means the semantics of `prep` are respected?

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: Sounds reasonable

wsmoses:
No your to_reac function does not achieve that. If you have a context of Tuple{Int, Int} this will not be converted unless you do `to_rarray(context; track_numbers=Number)`.

However, concurrently, most of the time you actually _want_ to partially evaluate integers in (e.g. for sizes/bounds/etc).

I think the more reasonable setup here is to not to_rarray the context, and instead add a similar warning to the one from reversediff:

```
These rules hold for the majority of backends, but there are some exceptions. The most important exception is [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl) and its taping mechanism, which is sensitive to control flow inside the function.
```

gdalle:
Being forced to keep the same context values makes preparation pretty much useless. I think I'd rather have us trace everything in the context, even if it means a slowdown in some cases. Will it lead to actual errors?

gdalle:
Or alternately we could restrict the kind of contexts we allow here

wsmoses:
Yes unnecessarily tracing objects can lead to errors that would fail to compile otherwise

wsmoses:
and it doesn't make it useless, it just means that the user is responsible for performing the to_rarray themselves for things that may change

gdalle:
> Yes unnecessarily tracing objects can lead to errors that would fail to compile otherwise

Can you give an example so that I wrap my mind around this?

> and it doesn't make it useless, it just means that the user is responsible for performing the to_rarray themselves for things that may change

That would be a Reactant-specific workaround, which doesn't fit other DI-supported backends. The whole point of DI is to enable easy backend switch, so I'd love to find a solution that doesn't expect users to wrap some of the arguments in Reactant-specific types when they want to switch to `AutoReactant`.

gdalle:
Besides, the problem is not specific to contexts: `x` itself can contain integers we don't necessarily want to track. And a preparation that can only be reused if nothing at all changes will only ever be used once anyway, so it is pointless.

Maybe DI could expose a function like `trace(a, backend)` or `translate(a, backend)` which takes care of populating values in the correct way for differentiation / Reactant compilation? I already use such a function internally anyway, especially in ForwardDiff and other operator overloading-based backends.

wsmoses:
```
julia> using Reactant; x = Reactant.to_rarray(ones(10)); s = Reactant.ConcreteRNumber(2); e = Reactant.ConcreteRNumber(5);

julia> f(x, s, e) = x[s:e]
f (generic function with 1 method)

julia> @jit f(x, s, e)
ERROR: TypeError: non-boolean (Reactant.TracedRNumber{Bool}) used in boolean context
Stacktrace:
  [1] getindex_linear
    @ ~/git/Reactant.jl/src/Indexing.jl:340 [inlined]
  [2] (::Nothing)(none::typeof(Reactant.TracedIndexing.getindex_linear), none::Reactant.TracedRArray{Float64, 1}, none::Reactant.TracedUnitRange{Reactant.TracedRNumber{Int64}})
    @ Reactant ./<missing>:0
  [3] getindex_linear
    @ ~/git/Reactant.jl/src/Indexing.jl:339 [inlined]
  [4] call_with_reactant(::Reactant.MustThrowError, ::typeof(Reactant.TracedIndexing.getindex_linear), ::Reactant.TracedRArray{…}, ::Reactant.TracedUnitRange{…})
    @ Reactant ~/git/Reactant.jl/src/utils.jl:0
  [5] getindex
    @ ~/git/Reactant.jl/src/Indexing.jl:75 [inlined]
  [6] (::Nothing)(none::typeof(getindex), none::Reactant.TracedRArray{Float64, 1}, none::Reactant.TracedUnitRange{Reactant.TracedRNumber{Int64}})
    @ Reactant ./<missing>:0
  [7] getindex
    @ ~/git/Reactant.jl/src/Indexing.jl:75 [inlined]
  [8] call_with_reactant(::Reactant.MustThrowError, ::typeof(getindex), ::Reactant.TracedRArray{Float64, 1}, ::Reactant.TracedUnitRange{Reactant.TracedRNumber{Int64}})
    @ Reactant ~/git/Reactant.jl/src/utils.jl:0
  [9] f
    @ ./REPL[5]:1 [inlined]
 [10] (::Nothing)(none::typeof(f), none::Reactant.TracedRArray{Float64, 1}, none::Reactant.TracedRNumber{Int64}, none::Reactant.TracedRNumber{Int64})
    @ Reactant ./<missing>:0
 [11] TracedUnitRange
    @ ~/git/Reactant.jl/src/Types.jl:108 [inlined]
 [12] TracedUnitRange
    @ ~/git/Reactant.jl/src/TracedRange.jl:124 [inlined]
 [13] Colon
    @ ~/git/Reactant.jl/src/TracedRange.jl:181 [inlined]
 [14] f
    @ ./REPL[5]:1 [inlined]
 [15] call_with_reactant(::typeof(f), ::Reactant.TracedRArray{Float64, 1}, ::Reactant.TracedRNumber{Int64}, ::Reactant.TracedRNumber{Int64})
    @ Reactant ~/git/Reactant.jl/src/utils.jl:0
 [16] make_mlir_fn(f::typeof(f), args::Tuple{…}, kwargs::@NamedTuple{}, name::String, concretein::Bool; toscalar::Bool, return_dialect::Symbol, args_in_result::Symbol, construct_function_without_args::Bool, do_transpose::Bool, input_shardings::Nothing, output_shardings::Nothing, runtime::Val{…}, verify_arg_names::Nothing, argprefix::Symbol, resprefix::Symbol, resargprefix::Symbol, num_replicas::Int64, optimize_then_pad::Bool)
    @ Reactant.TracedUtils ~/git/Reactant.jl/src/TracedUtils.jl:345
 [17] make_mlir_fn
    @ ~/git/Reactant.jl/src/TracedUtils.jl:275 [inlined]
 [18] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::typeof(f), args::Tuple{…}, compile_options::CompileOptions, callcache::Dict{…}, sdycache::Dict{…}, sdygroupidcache::Tuple{…}; fn_kwargs::@NamedTuple{}, backend::String, runtime::Val{…}, legalize_stablehlo_to_mhlo::Bool, client::Reactant.XLA.PJRT.Client, kwargs::@Kwargs{})
    @ Reactant.Compiler ~/git/Reactant.jl/src/Compiler.jl:1608
 [19] compile_mlir!
    @ ~/git/Reactant.jl/src/Compiler.jl:1570 [inlined]
 [20] compile_xla(f::Function, args::Tuple{…}; before_xla_optimizations::Bool, client::Nothing, serializable::Bool, kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/git/Reactant.jl/src/Compiler.jl:3516
 [21] compile_xla
    @ ~/git/Reactant.jl/src/Compiler.jl:3488 [inlined]
 [22] compile(f::Function, args::Tuple{…}; kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/git/Reactant.jl/src/Compiler.jl:3592
 [23] top-level scope
    @ ~/git/Reactant.jl/src/Compiler.jl:2661
Some type information was truncated. Use `show(err)` to see complete types.

julia> @jit f(x, 2, 5)
4-element ConcretePJRTArray{Float64,1}:
 1.0
 1.0
 1.0
 1.0
```

wsmoses:
and I don't think this is terribly reactant-specific. The same core issue here equally applies to reversediff compiled [where the context will equally be baked it]. Just because reactant also has a way to circumvent the problem in special cases shouldn't mean it is treated differently here.

wsmoses:
> and I don't think this is terribly reactant-specific. The same core issue here equally applies to reversediff compiled [where the context will equally be baked it]. Just because reactant also has a way to circumvent the problem in special cases shouldn't mean it is treated differently here.

bumping this @gdalle are you okay not to trace the contexts?

gdalle:
Not really. Many use cases of DI that I can think of require changing contexts, and Reactant has to be relevant for these cases too. Re-compiling the derivatives for each context changes is very impractical. On the other hand, asking users to pass `RArray`s instead of their normal arguments might make other backends fail, so the code on which DI runs is no longer fully generic, and I want to avoid that too. It would be like asking ForwardDiff users to pass arrays of `Dual` numbers.
Besides, that problem is not specific to contexts: what is stopping `x` itself (the active argument) from containing scalars that users may or may not want to trace?

I don't have a lot of bandwith these days, but I think the right solution might be to expose something like `DI.to_reactant`, telling users that the function will be called on every argument before Reactant compilation. That way, if they want to enforce specific tracing behavior, they can wrap their argument in a custom type and overload `to_reactant`, but it doesn't force them to

wsmoses:
In order to be differentiated the data must be a reactant array. If we assume that DI only officially supports array inputs this is fine

gdalle:
> In order to be differentiated the data must be a reactant array. If we assume that DI only officially supports array inputs this is fine

"array" as in "`RArray` only" or as in "any nested struct of `RArray`s?

gdalle:
By the way, [this Reactant issue](https://github.com/EnzymeAD/Reactant.jl/issues/1875) prevented me from testing `DI.Cache` here, if you happen to have a quick fix lying around. I can also modify the DI EnzymeExt source code if this is expected behavior

gdalle:
@wsmoses we're still missing tests but I added detailed documentation on the behavior that you wanted us to enforce. Can you review it to make sure it is sensible?

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -19,6 +19,8 @@ We support the following dense backend choices from [ADTypes.jl](https://github.
 - [`AutoTracker`](@extref ADTypes.AutoTracker)
 - [`AutoZygote`](@extref ADTypes.AutoZygote)
 
+In addition, we provide experimental support for [`AutoReactant`](@extref ADTypes.AutoReactant), sofar only for [`gradient`](@ref) and its variants.
```
wsmoses, line 22: Can we remove the experimental word, but just say the incomplete implementation?

wsmoses: For example it's better supported here than Diffractor listed above as fully supported

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,78 @@
...
+    compiled_value_and_gradient::CVG
+    compiled_value_and_gradient!::CVG!
+end
+
+function DI.prepare_gradient_nokwarg(
+        strict::Val, f::F, rebackend::AutoReactant, x, contexts::Vararg{DI.Context, C}
+    ) where {F, C}
+    _sig = DI.signature(f, rebackend, x; strict)
+    backend = rebackend.mode
+    xr = x isa ConcreteRArray ? nothing : ConcreteRArray(x)
```
wsmoses, line 16: If desired you could also do a check for abstract number and select ConcreteRNumber as a result

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,78 @@
...
+    return gr
+end
+
+function DI.value_and_gradient(
+        f::F, prep::ReactantGradientPrep, rebackend::AutoReactant, x, contexts::Vararg{DI.Context, C}
+    ) where {F, C}
+    DI.check_prep(f, prep, rebackend, x)
+    backend = rebackend.mode
+    xr = isnothing(prep.xr) ? x : copyto!(prep.xr, x)
+    contextsr = map(_to_reactant, contexts)
```
wsmoses, line 51: If you're going to use this design you need three functions to_reactant, copyto and copyfrom 

Honestly I wouldn't put this functionality into DI atm, users can already do this before the DI call and it's easier to add functions than remove them later

wsmoses (approved):

DifferentiationInterface/docs/src/explanation/backends.md
gdalle: Diffractor is listed in the README as broken, but sure

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: > Honestly I wouldn't put this functionality into DI atm, users can already do this before the DI call and it's easier to add functions than remove them later

The point of putting it in DI is that the _same code_ will work with different backends. If you force people to convert their arguments to Reactant types _before_ the DI call, these types may interact badly with some other autodiff backends, leading to code that is Reactant-only.
By the same token, I don't ask users to create their own `ForwardDiff.Dual` storage, or their own symbolic numbers, etc.

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: DI doesn't support `gradient` with number inputs for all backends (ForwardDiff will fail for instance), so the official API guarantee is for `x::AbstractArray`. Of course, reverse-mode backends are more generic than that, but I don't think it's bad to keep that restriction for now

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: > If you're going to use this design you need three functions to_reactant, copyto and copyfrom

Only for performance reasons, right? Since you said that Reactant's main use would be with non-traced `Constant`s, the conversion will be a no-op anyway in such cases.
I'd be in favor of trying it this way and then thinking about copy utilities as a second step.

DifferentiationInterface/docs/src/explanation/backends.md
gdalle: Another argument for the "experimental" support is protecting us if we decide that we made an API mistake, because breaking an "experimental" feature is arguably ok within SemVer

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
gdalle: Also, it's difficult to decide whether a type is mutable (hence amenable to `copy!`) or not. For instance, https://github.com/EnzymeAD/Enzyme.jl/issues/2587 doesn't have a satisfying solution yet

wsmoses: I agree with that for the differentiated argument (which is what is done elsewhere too), but this is already extremely backend specific and seems likely to confuse people wrt the reactant.to_rarray and such.

Given the discussion above that the absence of these as a default is not going to be likely to cause problems, and it's easier to add APIs later without breakage than remove them, I think it's better to start simple and add this later if it becomes relevant

gdalle: I strongly disagree and I would ask you to trust my gut on this one, based on DI usage I'm seeing in the wild. While scenarios that you care about mostly seem to include non-traceable constants, there are plenty of other cases where users will need the flexibility of picking which contexts they trace and how. If we don't give it to them, they may never even try the switch to Reactant, which I think would be a poor outcome.

Putting this behind a big "experimental: expect breaking stuff" warning, at least for the upcoming release, keeps us free to modify the API if we get the design wrong the first time. In addition, I've been playing with the idea of a similar "translation" API for other backends, so it may not end up being backend-specific after all. I could rename this into `translate(backend, arg)` and let users customize it also for dual or symbolic numbers depending on their needs.

wsmoses: I hear you, but I strongly disagree as well.

As to your specific point:
"they may never even try the switch to Reactant, which I think would be a poor outcome"

That is totally okay. Usage within DI is meant as a quick validation and ease of use. In most cases reactant will be called outside of any derivative calls, not the inside of like here.

More importantly, if we start small and simple here, we can expand based off of actual usage instead of pre-supposing usage. It also keeps the API a lot cleaner. However you can't as easily un-ring the bell.

gdalle: We can't expand based on actual usage if we only allow the narrowest possible usage you can think of. You know existing Reactant users, but I know existing DI users, which are very varied, and I also want to open the doors to applications and users that neither of us knows about yet.

And let me push back on DI mostly being used as quick prototyping, a temporary stepping stone. No doubt that some people using DI could get even more performance by switching to native Enzyme. But many of them won't, precisely because it's a hassle to refactor one more time. The same will likely happen with Reactant: sure, compiling the outer loop would be best, but I'm willing to bet plenty of people will be happy with just accelerating their inner gradient, and it will already change their life for the better.

A clean API is not useful if you cannot do anything with it. Only allowing preparation-time constants and no cache completely strips DI of its substance, and `AutoReactant` of its usefulness.
I'm the one who will document, test and maintain the API, so I'm willing to take responsibility for how clean it is. I haven't had too many complaints about DI being unclean so far, and even if I do get them, changing an experimental API is _by definition_ not breaking. That's why I intend to keep the warning, so we can let people experiment without being afraid to backtrack if we got it wrong.

gdalle:
Some examples of DI users where `Constant` changes between preparation and execution:

- AdvancedVI.jl: minibatch samples passed as `Constant` (see [here](https://github.com/TuringLang/AdvancedVI.jl/blob/17d61f9299f93d87c0eeadf9b60c87f8c44b1d11/src/algorithms/scoregradelbo.jl#L106-L112) and [here](https://github.com/TuringLang/AdvancedVI.jl/blob/17d61f9299f93d87c0eeadf9b60c87f8c44b1d11/src/AdvancedVI.jl#L51-L58))
- ImplicitDifferentiation.jl: other argument of the implicit function theorem passed as `Constant` to get partial derivatives (see [here](https://github.com/JuliaDecisionFocusedLearning/ImplicitDifferentiation.jl/blob/5380155fe628c44cc9dbd35fcec7afe80a2d3728/src/execution.jl#L52-L58) and [here](https://github.com/JuliaDecisionFocusedLearning/ImplicitDifferentiation.jl/blob/5380155fe628c44cc9dbd35fcec7afe80a2d3728/src/execution.jl#L160-L170))
- MarginalLogDensities.jl: some "data" passed as `Constant` (see [here](https://github.com/ElOceanografo/MarginalLogDensities.jl/blob/34f6f0baae616ed5ca0d7608348f628b87c9b49f/src/MarginalLogDensities.jl#L203-L217))
- ModelPredictiveControl.jl: controls `u` passed as `Constant` (see [here](https://github.com/JuliaControl/ModelPredictiveControl.jl/blob/2730e8b10f7596f075c8ffbe372c939e63e79524/src/estimator/kalman.jl#L1182-L1191))
- ModelingToolkit.jl: time `t` (among other things) passed as `Constant` (see [here](https://github.com/SciML/ModelingToolkit.jl/blob/5a49e934941c775c44903c55cf1508e56cc916ee/src/linearization.jl#L318-L319))
- NLSolversBase.jl (and thus Optim.jl): Lagrange multipliers passed as `Constant` (see [here](https://github.com/JuliaNLSolvers/NLSolversBase.jl/blob/7c56890c1502cc2c73c72c0ffb1540773cf9c0fe/src/objective_types/constraints.jl#L190-L193))
- All of SciML: parameters `p` passed as `Constant` (not really sure what's in there)

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
```diff
@@ -0,0 +1,78 @@
+struct ReactantGradientPrep{SIG, XR, GR, CG, CG!, CVG, CVG!} <: DI.GradientPrep{SIG}
+    _sig::Val{SIG}
+    xr::XR
+    gr::GR
+    compiled_gradient::CG
+    compiled_gradient!::CG!
```
wsmoses, line 6: can we have different prep objects for each of the compiled variants. Reason being that one may compile whereas the other may not.

gdalle: I'm not sure what you mean here. You want to explore all 2^4 combinations of compiled/non-compiled operator variants?

wsmoses: I mean that we should have distinct

```
struct ReactantGradientPrep
    compiled_gradient::CG
end

struct ReactantGradient!Prep
    compiled_gradient!::CG!
end

struct ReactantValueAndGradient!Prep
    compiled_value_and_gradient!::CG!
end

....
```

perhaps they can be templated or anything else, but they should be distinct (and therefore prepare should only compile for the one that will be used)

gdalle: That is not compatible with DI's API. A preparation result must allow calling all four variants of an operator.
Would it speed things up if the compiled versions built off one another?

wsmoses: Nope, and rip that seems like a design limitation of DI.

Is it something that's fixable in time?

gdalle: Just because you don't like it doesn't automatically mean it is a design limitation, it could also just be a design decision you disagree with 😉 In fact, it is completely consistent with the way most backends work, otherwise I would have done it differently. And it is rather convenient for users having 1 preparation for 4 variants.

It is not fixable without a breaking release, which I'm fairly reluctant to do given that we have 1000 indirect dependents.
Even if we did introduce variants like `prepare_gradient!` or `prepare_value_and_gradient`, which we can do fairly easily, the basic `prepare_gradient` would still need to prepare for all 4 variants.

gdalle: See https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/685 to continue this design discussion (I opened this issue a year ago ^^)

wsmoses: gentle follow up here, from the discussion on slack, it seemed like the community was responsive to this?

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
wsmoses: I remember we had come to a conclusion on this on slack, but unfortunately it appears lost to the slack-hole. what are the next steps here?

DifferentiationInterface/src/misc/reactant.jl
```diff
@@ -0,0 +1,84 @@
+"""
+!!! tip "DI-specific information"
+    This part of the docstring is related to the use of `AutoReactant` inside [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl), or DI for short.
+    Reactant's tutorial on [partial evaluation](https://enzymead.github.io/Reactant.jl/stable/tutorials/partial-evaluation) is useful reading to understand what follows.
+
+The `AutoReactant` backend inside DI imposes the following restrictions / assumptions:
+
+- The only supported operator (at the moment) is `DI.gradient` (along with its variants).
+- The input `x` must be an `AbstractArray` such that `Reactant.ConcreteRArray(x)` is well-defined.
+- By default, contexts such as `DI.Constant` and `DI.Cache` will be partially evaluated inside the compiled differentiation operator at preparation time. This means that the context value provided at preparation will be reused at every subsequent execution, while the context value provided at execution will be ignored. In particular, `DI.Cache` contexts will usually error and `DI.Constant` contexts will be frozen to one value.
```
penelopeysm, line 10: The extensibility of `to_reactant` is worth keeping because users might have weird structs or types where they want to very specifically control tracing. However, my guess is that most real usage is going to be with `AbstractArray{<:Number}` or `Number`. And with the current state of this PR, regardless of whether you define `Constant` to trace or not to trace, if a user wants to do the _other_ thing, they have to not only define their own `AbstractArray` wrapper, overload `DI.to_reactant`, and additionally modify the rest of their code to unwrap that wrapper. Which is quite a pain, not least for something that's notionally supposed to abstract away (some of) the difficulty of using Reactant.

I more or less think that the only acceptable solution to this is a third context, let's say `ConstantConstant` i.e., a value that is formally fixed at preparation time. Naturally, for Reactant, we don't trace `ConstantConstant`s, but we do trace `Constant`s.

For all other backends, `ConstantConstant` can have the same behaviour as `Constant`, which means that the precise semantics of `ConstantConstant` are "changing the value of that at execution time _might_ work 'as intended' but is not guaranteed to give you correct answers so please don't do it". Then the semantics of `Constant` remain the same: "changing the value of this at execution time _will_ work".

An alternative is for DI to define its own Array wrappers which have the correct semantics, so something like

```julia
struct PrepTimeConstant{D}
    data::D
end

DI.to_reactant(ptc::PrepTimeConstant{<:AbstractArray}) = ptc.data
DI.to_reactant(arr::AbstractArray) = Reactant.to_rarray(arr)
```

So `Constant(PrepTimeConstant(data))` gives you a true preparation-time constant, analogous to `ConstantConstant` above, and `Constant(data)` gives a traced array. But I'm not sure I like the "magic" auto-unwrapping for the user.

DifferentiationInterface/ext/DifferentiationInterfaceReactantExt/onearg.jl
penelopeysm: Personally for this one I'd be fine with as is. It's definitely not optimal but imo the main question is what do you want to achieve with the integration. Given that the aims of Reactant and DI are not entirely aligned, I don't think it's likely that the Reactant+DI combination will ever be optimal so I don't think that perfection should be the aim here - I think it would already be a win to get something that works "well enough" for "most' people, and then if people want to (say) squeeze out additional perf by not compiling things they don't need, then they can go the old route of having a custom ReactantExt and bypassing DI.

gdalle:
Thanks @penelopeysm for your feedback! I have launched an agent to work on this and submit a detailed plan over in #265, let's see what Claude can come up with to remain as flexible as possible design-wise. It won't be optimal yet, but I agree that it doesn't have to be optimal to be useful.

wsmoses (changes requested):

PR #932, opened 2025-11-21
fix: improve wrong-mode pushforward/pullback

gdalle:


.github/workflows/Test.yml
```diff
@@ -166,7 +166,7 @@ jobs:
       actions: write
       contents: read
     strategy:
-      fail-fast: true
+      fail-fast: false # TODO: toggle
```
gdalle:
```suggestion
      fail-fast: true # TODO: toggle
```

PR #940, opened 2025-12-03
feat: add support for HyperHessian backend

KristofferC:
HyperHessian (https://github.com/KristofferC/HyperHessians.jl) is a forward mode AD package that specializes in second-order derivatives (Hessians). I initially made it like 4 years ago as an experiment to see what would happen if you made an AD package using hyper duals over nested duals for second order derivative but only recently freshened it up to be a proper package. The hyper dual number approach does have some extra flexibility over nested duals, which allows it to squeeze out some more performance over e.g. ForwardDiff:

### [Hessian](https://github.com/KristofferC/HyperHessians.jl/blob/master/benchmark/fdiff.jl): 
| Function      | input length | Time ForwardDiff | Time HyperHessians | Speedup |
| ------------- | ------------ | ---------------- | ----------------- | --------|
| `ackley` | 1 | 50.215 ns | 34.872 ns | 1.4
| `ackley` | 8 | 753.440 ns | 292.278 ns | 2.6
| `ackley` | 128 | 2.026 ms | 494.049 μs | 4.1
| `ackley` | 1024 | 1.062 s | 236.522 ms | 4.5
| `rosenbrock_1` | 1 | 24.240 ns | 7.756 ns | 3.1
| `rosenbrock_1` | 8 | 908.258 ns | 298.307 ns | 3.0
| `rosenbrock_1` | 128 | 3.163 ms | 576.477 μs | 5.5
| `rosenbrock_1` | 1024 | 1.704 s | 280.535 ms | 6.1

### [Hvp](https://github.com/KristofferC/HyperHessians.jl/blob/master/benchmark/hvp.jl)

| Function      | input length | Time ForwardDiff | Time HyperHessians | Speedup |
| ------------- | ------------ | ---------------- | ------------------ | ------- |
| `rosenbrock_1` | 8 | 198.708 ns | 127.058 ns | 1.6 |
| `rosenbrock_1` | 16 | 893.167 ns | 555.688 ns | 1.6 |
| `rosenbrock_1` | 128 | 46.955 μs | 32.693 μs | 1.4 |
| `rosenbrock_1` | 1024 | 2.897 ms | 2.089 ms | 1.4 |
| `ackley` | 8 | 297.756 ns | 124.167 ns | 2.4 |
| `ackley` | 16 | 1.138 μs | 464.128 ns | 2.5 |
| `ackley` | 128 | 46.364 μs | 24.874 μs | 1.9 |
| `ackley` | 1024 | 2.807 ms | 1.559 ms | 1.8 |

(note that I am not claiming that it is impossible to improve ForwardDiff further w.r.t hessians, I am just saying that these are the numbers as of right now)

The package is not registered yet, but I thought I would open this for some feedback anyway. Note that some of the stuff in here was written with an LLM because figuring out exactly what to overload, etc, was tricky. Hopefully, it is a good starting point to iterate on. I also haven't checked that any overhead of DI vs raw HyperHessians is negligible yet.


gdalle:
Thanks for opening a PR!
If you're going to iterate a lot, can you comment out the irrelevant tests in Test.yml (remove everything related to DIT, and toggle a comment on everything that is not your backend)? The CI suite of DI is _extremely_ expensive

gdalle:
As far as knowing what to overload, maybe this page can help: https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/dev/contributing/

KristofferC:
> Thanks for opening a PR!
> If you're going to iterate a lot, can you comment out the irrelevant tests in Test.yml (remove everything related to DIT, and toggle a comment on everything that is not your backend)? The CI suite of DI is extremely expensive

I did so, feel free to cancel the running tests on previous commit

gdalle:
You can also comment out the whole `test-DI-Core` and `test-DIT` jobs.
I wish I knew a smarter way to indicate which test sets to run, but it's hard to anticipate which parts of the codebase need to be re-tested based on a given change.

KristofferC:
At least the tests for the HyperHessian backend seems to pass.

KristofferC:
What's my next step? Do I need to add `AutoHyperHessian` to ADTypes.jl?

gdalle:
Maybe give me a few days to review this first? Also, it would be the first backend that implements only a few first-order operations, so I'm not sure how to document that

KristofferC:
> Also, it would be the first backend that implements only a few first-order operations, so I'm not sure how to document that

 Maybe the table of supported features is enough?

I could add gradient, etc, but the hyperdual number would then degenerate to a normal dual number, and it would feel a bit silly to just re-implement things that are functionally identical to ForwardDiff.

gdalle:
It's not a table of supported features, it's a table of which operators are implemented natively by the backend. For instance, Mooncake only implements a pullback, but thanks to DI's internal machinery you get a Jacobian operator for free. With this package, there is no pullback or pushforward so that machinery will fail

gdalle:
It's not necessarily a bad thing, just a new thing

DifferentiationInterface/src/misc/autohyperhessians.jl
```diff
@@ -0,0 +1,16 @@
+"""
+    AutoHyperHessians(; chunksize = nothing)
+
+Lightweight ADTypes backend tag for HyperHessians. The `chunksize` keyword can
+be set to a positive `Int` to override HyperHessians' chunk heuristic; `nothing`
+lets HyperHessians choose.
+"""
+struct AutoHyperHessians{CS} <: ADTypes.AbstractADType
```
gdalle: Just adding a comment to remember we shouldn't merge this, since it belongs in ADTypes

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -69,6 +71,7 @@ Moreover, each context type is supported by a specific subset of backends:
 | `AutoFiniteDifferences`    | ✅                  | ✅               |
 | `AutoForwardDiff`          | ✅                  | ✅               |
 | `AutoGTPSA`                | ✅                  | ❌               |
+| `AutoHyperHessians`        | ✅                  | ✅               |
```
gdalle: I don't think you support caches, that would require dealing with hyperdual element types

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,259 @@
+module DifferentiationInterfaceHyperHessiansExt
+
+import DifferentiationInterface as DI
+import .DI: AutoHyperHessians
+using ADTypes: ForwardMode
+using HyperHessians:
```
gdalle, line 5: Are all of these public, as in, protected from API changes in minor or patch versions?

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,259 @@
...
+    hessian,
+    hvp,
+    hvp!,
+    hvp_gradient_value,
+    hvp_gradient_value!
+
+## Traits
+DI.check_available(::DI.AutoHyperHessians) = true
+DI.inplace_support(::DI.AutoHyperHessians) = DI.InPlaceSupported()
+DI.hvp_mode(::DI.AutoHyperHessians) = DI.ForwardOverForward()
```
gdalle:
```suggestion
```
This is deduced automatically from the `mode`

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,259 @@
...
+DI.mode(::DI.AutoHyperHessians) = ForwardMode()
+
+chunk_from_backend(backend::DI.AutoHyperHessians, x) =
+    isnothing(backend.chunksize) ? Chunk(x) : Chunk{backend.chunksize}()
+chunk_from_backend(backend::DI.AutoHyperHessians, N::Integer, ::Type{T}) where {T} =
+    isnothing(backend.chunksize) ? Chunk(pickchunksize(N, T), T) : Chunk{backend.chunksize}()
+
+function DI.pick_batchsize(backend::DI.AutoHyperHessians, x::AbstractArray)
```
gdalle, line 29: You're missing the same function for integer arguments instead of arrays

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,259 @@
...
+
+function DI.second_derivative(
+        f,
+        prep::HyperHessiansSecondDerivativePrep,
+        backend::DI.AutoHyperHessians,
+        x::Number,
+        contexts::Vararg{DI.Context, C},
+    ) where {C}
+    DI.check_prep(f, prep, backend, x, contexts...)
+    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
```
gdalle: This will only work with `Constant` contexts (see the ForwardDiff extension for how I deal with `Cache`)

gdalle (changes requested):
Overall this seems reasonable, the main decision to make now is whether you want to support `Cache` or not. I don't think it is crucial at first glance.

Could you maybe comment on https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/828 to see if I missed anything vis-a-vis creating a new backend? I should probably turn that issue into docs but I'd be curious about your (LLM-enhanced) experience

gdalle:
Also, the tests are failing

DifferentiationInterface/docs/src/explanation/backends.md
KristofferC: Ok, I thought that was just for example the `HessianConfig` etc.

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
KristofferC: Yes

DifferentiationInterface/test/Back/HyperHessians/test.jl
```diff
@@ -0,0 +1,37 @@
...
+import DifferentiationInterfaceTest as DIT
+using ExplicitImports
+using HyperHessians
+using Test
+
+check_no_implicit_imports(DifferentiationInterface)
+
+backends = [
+    DI.AutoHyperHessians(),
+    DI.AutoHyperHessians(; chunksize = 4),
```
gdalle, line 14: I'm surprised tests pass without implementing `threshold_batchsize`. Maybe all second-order scenarios have length >= 4, but we should still add that function to the extension.

DifferentiationInterface/test/Back/HyperHessians/test.jl
```diff
@@ -0,0 +1,37 @@
...
+test_differentiation(
+    backends, scenarios;
+    excluded = excluded_ops, logging = LOGGING,
+)
+
+test_differentiation(
+    DI.AutoHyperHessians(), scenarios;
+    correctness = false,
+    type_stability = safetypestab(:prepared),
+    excluded = excluded_ops,
```
gdalle:
```suggestion
    excluded = FIRST_ORDER,
```

DifferentiationInterface/test/Back/HyperHessians/test.jl
```diff
@@ -0,0 +1,37 @@
...
+    DI.AutoHyperHessians(; chunksize = 4),
+]
+
+for backend in backends
+    @test DI.check_available(backend)
+    @test DI.check_inplace(backend)
+end
+
+excluded_ops = [:pushforward, :pullback, :jacobian, :derivative, :gradient]
+
```
gdalle:
```suggestion
```

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -12,6 +12,7 @@ We support the following dense backend choices from [ADTypes.jl](https://github.
 - [`AutoFiniteDifferences`](@extref ADTypes.AutoFiniteDifferences)
 - [`AutoForwardDiff`](@extref ADTypes.AutoForwardDiff)
 - [`AutoGTPSA`](@extref ADTypes.AutoGTPSA)
+- [`AutoHyperHessians`](https://github.com/KristofferC/HyperHessians.jl)
```
gdalle: Adding a comment to change this once the ADTypes exists

DifferentiationInterface/test/Back/HyperHessians/test.jl
```diff
@@ -0,0 +1,37 @@
...
+scenarios = default_scenarios(; include_constantified = true)
```
gdalle:
```suggestion
scenarios = default_scenarios(; include_constantified = true, include_cachified=true)
```

gdalle: This is important to test caches

gdalle (changes requested):
The docs failure is expected, since the docstring of the backend will be in ADTypes instead

KristofferC:
Updated

KristofferC:
Any guide for what I should do as a next step here?

gdalle:
I solemnly swear that this will be reviewed by next Monday

DifferentiationInterface/docs/src/explanation/backends.md
gdalle: No, in DI parlance the `Config` object is called a "preparation". What I mean by "cache" is explained here: https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/explanation/arguments/#Contexts
Do you think you want to support it?

DifferentiationInterface/docs/src/explanation/backends.md
gdalle: @KristofferC can you open the ADTypes PR?

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,314 @@
...
+    hessian_gradient_value!,
+    hessian,
+    hvp,
+    hvp!,
+    hvp_gradient_value,
+    hvp_gradient_value!
+
+## Traits
+DI.check_available(::DI.AutoHyperHessians) = true
+DI.inplace_support(::DI.AutoHyperHessians) = DI.InPlaceSupported()
```
gdalle: Is that true? This only makes sense for "something-to-vector" functions `f!(y, x)`, but I don't think HyperHessians deals with those?

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,314 @@
...
+
+chunk_from_backend(backend::DI.AutoHyperHessians, x) =
+    isnothing(backend.chunksize) ? Chunk(x) : Chunk{backend.chunksize}()
+chunk_from_backend(backend::DI.AutoHyperHessians, N::Integer, ::Type{T}) where {T} =
+    isnothing(backend.chunksize) ? Chunk(pickchunksize(N, T), T) : Chunk{backend.chunksize}()
+
+function DI.pick_batchsize(backend::DI.AutoHyperHessians, x::AbstractArray)
+    B = chunksize(chunk_from_backend(backend, x))
+    return DI.BatchSizeSettings{B}(length(x))
+end
```
gdalle, line 32: Is there a way to make this type-stable for StaticArrays?

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,314 @@
...
+function translate_prepared(
+        contexts::NTuple{C, DI.Context}, prep_contexts::NTuple{C, Any}
+    ) where {C}
+    new_contexts = map(contexts, prep_contexts) do c, pc
+        _translate_prepared(c, pc)
+    end
+    return new_contexts
+end
+
+## Second derivative (scalar input)
```
gdalle, line 74: Strictly speaking it can also have a vector output, does your package handle that?

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
```diff
@@ -0,0 +1,314 @@
...
+        x::Number,
+        contexts::Vararg{DI.Context, C},
+    ) where {C}
+    DI.check_prep(f, prep, backend, x, contexts...)
+    contexts_prepared = translate_prepared(contexts, prep.contexts_prepared)
+    fc = DI.fix_tail(f, contexts_prepared...)
+    return hessian(fc, x)
+end
+
+function DI.second_derivative!(
```
gdalle, line 102: This won't be tested if you can only handle scalar outputs

gdalle (changes requested):

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
KristofferC: This is the same as ForwardDiff so for now perhaps it is ok to have it like this?

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
KristofferC: For scalar inputs, HH works with any output (`num_to_vec` and `num_to_mat` test pass).

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
KristofferC: yes

DifferentiationInterface/ext/DifferentiationInterfaceHyperHessiansExt/DifferentiationInterfaceHyperHessiansExt.jl
KristofferC: i changed this

DifferentiationInterface/test/Back/HyperHessians/test.jl
KristofferC: implemented

DifferentiationInterface/docs/src/explanation/backends.md
KristofferC: Done in https://github.com/SciML/ADTypes.jl/pull/146

KristofferC:
I think I have addressed the review.

ADTypes PR at https://github.com/SciML/ADTypes.jl/pull/146 (tests here wont pass until it, but it passes locally with it devved).

gdalle:
Thanks! Any idea why code coverage is missing?

KristofferC:
Maybe didnt run because test didn't pass if that is what you mean. Still need a new ADTypes for that.

gdalle:
I just tagged and released your new ADTypes, can you update this PR accordingly, putting 1.22 as the compat lower bound on ADTypes? Thanks 😊

KristofferC:
Done

gdalle:
The error we see in CI is because you need to add and load JET explicitly to test type stability.
Can you merge main into this branch and do that small addition? After that we should be nearly there

gdalle:
I did it, let's see

KristofferC:
Sorry, I missed your earlier comment

DifferentiationInterface/Project.toml
```diff
@@ -15,6 +15,7 @@ Diffractor = "9f5e2b26-1114-432f-b630-d3fe2085c51c"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 FastDifferentiation = "eb9bf01b-bf85-4b60-bf87-ee5de06c00be"
+HyperHessians = "06b494a0-c8e0-40cc-ad32-d99506a00a6c"
```
gdalle, line 18: Can you bump DI's patch version?

DifferentiationInterface/docs/src/explanation/backends.md
```diff
@@ -69,6 +71,7 @@ Moreover, each context type is supported by a specific subset of backends:
 | `AutoFiniteDifferences`    | ✅                  | ✅               |
 | `AutoForwardDiff`          | ✅                  | ✅               |
 | `AutoGTPSA`                | ✅                  | ❌               |
+| `AutoHyperHessians`        | ✅                  | ❌               |
```
gdalle: It seems you now test `cachified` scenarios successfully?

DifferentiationInterface/Project.toml
gdalle: And add a Changelog entry?

gdalle (approved):
Two minor details and then we're good to go! Thanks

DifferentiationInterface/Project.toml
gdalle: done

DifferentiationInterface/docs/src/explanation/backends.md
gdalle: fixed

PR #926, opened 2025-11-19
chore: revamp testing CI with dedicated environments

gdalle:
- Split DI's testing CI between Core and Backends
- Introduce a dedicated environment for each backend, thus avoiding `Pkg` operations inside the test suite
- Remove Diffractor tests

PR #948, opened 2025-12-30
Fix inner preparation behavior for Mooncake

yebai:
See, https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3696647362

> In the case of Mooncake, we're dealing with source transformation, so DI.PrepareInnerSimple() is more appropriate, good job for finding that.

*Edit:* While no longer strictly necessary following PR [#926](https://github.com/chalk-lab/Mooncake.jl/pull/926) of Mooncake.jl, this remains useful as an optimisation.

### DIFF
--- a/DifferentiationInterface/Project.toml
+++ b/DifferentiationInterface/Project.toml
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.7.14"
+version = "0.7.13"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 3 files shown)


DifferentiationInterface/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 authors = ["Guillaume Dalle", "Adrian Hill"]
-version = "0.7.13"
+version = "0.7.14"
```
```suggestion
version = "0.7.13"
```
I usually do version updates all at once

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/DifferentiationInterfaceMooncakeExt.jl
```diff
@@ -34,6 +34,7 @@ using Mooncake:
 const AnyAutoMooncake{C} = Union{AutoMooncake{C}, AutoMooncakeForward{C}}
 
 DI.check_available(::AnyAutoMooncake{C}) where {C} = true
+DI.inner_preparation_behavior(::AutoMooncakeForward) = DI.PrepareInnerSimple()
```
gdalle, line 37: This is only testable inside second-order tests, can you add them here now that https://github.com/chalk-lab/Mooncake.jl/pull/878 has been merged?

gdalle (changes requested):

gdalle (approved):
Thanks, this looks good! If the tests pass i'll merge

PR #954, opened 2026-01-22
fix: overloaded_input_type for one-element vector input

ErikQQY:
`overloaded_input_type` doesn't work for jacobian with one-element vector:

```julia
julia> backend = AutoForwardDiff()
AutoForwardDiff()

julia> DI.overloaded_input_type(prepare_jacobian(copy, backend, [1.0]))
ERROR: BoundsError: attempt to access 1-element Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(copy), Float64}, Float64, 1}} at index [2]
```

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/misc.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/misc.jl
@@ -28,5 +28,5 @@ DI.overloaded_input_type(prep::ForwardDiffTwoArgDerivativePrep) = typeof(prep.co
 DI.overloaded_input_type(prep::ForwardDiffGradientPrep) = typeof(prep.config.duals)
 
 ## Jacobian
-DI.overloaded_input_type(prep::ForwardDiffOneArgJacobianPrep) = typeof(prep.config.duals[2])
-DI.overloaded_input_type(prep::ForwardDiffTwoArgJacobianPrep) = typeof(prep.config.duals[2])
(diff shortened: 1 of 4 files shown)


gdalle:
Hi @ErikQQY, thanks for the suggested bug fix!
However, it turns out the issue runs deeper than that, and actually goes back to our first PR introducing this mechanism in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/672. We neglected several aspects back then:

- from `DerivativeConfig` in the two-argument case (in-place), we used to retrieve the type of `ydual` but we actually want the type of `xdual`.
- from `JacobianConfig` in the one-argument case (out-of-place), we tried to retrieve the type of `xduals` based on [this definition](https://github.com/JuliaDiff/ForwardDiff.jl/blob/71258ec502052fdf298279923652b34cc71f4a85/src/config.jl#L185-L187) but we ended up pulling `xduals[2]` because of [that definition](https://github.com/JuliaDiff/ForwardDiff.jl/blob/71258ec502052fdf298279923652b34cc71f4a85/src/config.jl#L159).
- some of the tests were wrong too

I tried to fix all of that, would you mind giving it a second, closer look, comparing it with https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/config.jl#L159?

Thanks

ErikQQY:
I think the current fix looks good now. The previous implementation mess up different cases😅

ErikQQY:
Can we get a new release with this patch? I need this fix for some new functionalities.

gdalle:
I tried to register a new version but Registrator didn't notice

PR #961, opened 2026-01-29
fix: upgrade Mooncake compat to v0.5

### LINKED ISSUE #642: Mooncake Backend doesn't handle functions with StaticArrays output
If a function returns a static array or vector, the AutoMooncake backend errors

```julia
using StaticArrays
using ForwardDiff
using Mooncake

function MWE(x::AbstractVector)

    z = SVector{3}(x.^2)

    return z

end

test = rand(3)

f_ad, df_ad = value_and_jacobian(
    MWE,
    AutoMooncake(; config=nothing),
    test
)

f_ad, df_ad = value_and_jacobian(
    MWE,
    AutoForwardDiff(),
    test
)
```

Gives the error 

> ERROR: MethodError: no method matching copyto!(::Mooncake.Tangent{@NamedTuple{data::Tuple{Float64, Float64, Float64}}}, ::SVector{3, Float64})
> The function `copyto!` exists, but no method is defined for this combination of argument types.
> 
> Closest candidates are:
>   copyto!(::IndexStyle, ::AbstractArray, ::IndexStyle, ::AbstractArray)
>    @ Base abstractarray.jl:1064
>   copyto!(::Zygote.Buffer, ::Any)
>    @ Zygote C:\Users\jmurp\.julia\packages\Zygote\nyzjS\src\tools\buffer.jl:54
>   copyto!(::PermutedDimsArray, ::AbstractArray)
>    @ Base permuteddimsarray.jl:295
>   ...
> 
> Stacktrace:
>  [1] copyto!!(dst::Mooncake.Tangent{@NamedTuple{data::Tuple{Float64, Float64, Float64}}}, src::SVector{3, Float64})
>    @ DifferentiationInterfaceMooncakeExt C:\Users\jmurp\.julia\packages\DifferentiationInterface\gSdHF\ext\DifferentiationInterfaceMooncakeExt\DifferentiationInterfaceMooncakeExt.jl:29
>  [2] value_and_pullback(::Function, ::DifferentiationInterfaceMooncakeExt.MooncakeOneArgPullbackPrep{…}, ::AutoMooncake{…}, ::Vector{…}, ::Tuple{…})
>    @ DifferentiationInterfaceMooncakeExt C:\Users\jmurp\.julia\packages\DifferentiationInterface\gSdHF\ext\DifferentiationInterfaceMooncakeExt\onearg.jl:35
>  [3] prepare_pullback(::Function, ::AutoMooncake{Nothing}, ::Vector{Float64}, ::Tuple{SVector{3, Float64}})
>    @ DifferentiationInterfaceMooncakeExt C:\Users\jmurp\.julia\packages\DifferentiationInterface\gSdHF\ext\DifferentiationInterfaceMooncakeExt\onearg.jl:22
>  [4] _prepare_jacobian_aux(::DifferentiationInterface.PushforwardSlow, ::DifferentiationInterface.BatchSizeSettings{…}, ::SVector{…}, ::Tuple{…}, ::AutoMooncake{…}, ::Vector{…})
>    @ DifferentiationInterface C:\Users\jmurp\.julia\packages\DifferentiationInterface\gSdHF\src\first_order\jacobian.jl:167
>  [5] prepare_jacobian(::typeof(MWE), ::AutoMooncake{Nothing}, ::Vector{Float64})
>    @ DifferentiationInterface C:\Users\jmurp\.julia\packages\DifferentiationInterface\gSdHF\src\first_order\jacobian.jl:108
>  [6] value_and_jacobian(::typeof(MWE), ::AutoMooncake{Nothing}, ::Vector{Float64})
>    @ DifferentiationInterface C:\Users\jmurp\.julia\packages\DifferentiationInterface\gSdHF\src\fallbacks\no_prep.jl:60
>  [7] top-level scope
>    @ c:\Users\jmurp\.julia\dev\SatelliteToolboxGravityModels\test\differentiability.jl:69
> Some type information was truncated. Use `show(err)` to see complete types.

gdalle:
- [x] Bump Mooncake compat to v0.5.1
- [x] Take `Config` as a whole instead of parsing it
- [x] Remove useless `_righttype` caches
- [x] Add tests with StaticArrays for new friendly tangents mode

Supersedes #960

See also https://github.com/chalk-lab/Mooncake.jl/issues/979

Note that this still uses internal conversion machinery from Mooncake


Technici4n:
This is OK but you should at least carry over the `dy_righttype` changes, it never made sense to do that and breaks friendly tangents.

Technici4n (approved):
I guess this is fine if you just want to get this out of the door!

gdalle:
What do you mean by "never really made sense to do that"?

gdalle:
Perhaps we should just merge the Mooncake PR and do everything at once

Technici4n:
What I mean is that `_copy_output` is just a copy, it doesn't change the object's type, so I don't understand why it is there unless the intention is specifically to make a copy. This PR is fine and should fix Mooncake CI (including docs).

AstitvaAggarwal:
looks good, thanks!

gdalle:
Let's just get https://github.com/chalk-lab/Mooncake.jl/pull/980/changes merged and we'll clean up the DI bindings as we bump our compat bound to Mooncake v0.5.1

gdalle:
So I tracked down the reason behind the existence of `dx_righttype` and `dy_righttype` to https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/513. At first, it was just a way to safeguard against the user providing, say, `x::Vector{Float64}` and `dx::Vector{Float32}`, similar to what is done in the Enzyme extension. For lack of friendly tangents infrastructure, the best one could do is make sure that the primal type was equal to the tangent type, and this would work for simple numbers and arrays (which is all I test on).

Initially, this used the handrolled function
```julia
copyto!!(dst::Number, src::Number) = convert(typeof(dst), src)
copyto!!(dst, src) = copyto!(dst, src)
```
But then, in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/809, I replaced this function with the one from Mooncake, neglecting the fact that it doesn't accept inputs of different (el)types. Which means the functionality has been essentially useless for a while and we should remove it during the switch to Mooncake 0.5, now that friendly tangents take care of conversion.

Technici4n:
I see, that makes sense, thanks for investigating. ^^

gdalle:
Tests currently fail in forward mode with friendly tangents and in-place function because the `dy` provided by the user is not necessarily the one that gets modified. I think I have to use the same trick as for reverse mode and define a new
```julia
f!_with_return(y, x) = begin
    f!(y, x)
    return y
end
```

Technici4n (approved):

sunxd3:
Thanks for the work!

Issue #806, opened 2025-05-26
Enhancing `DifferentiateWith` Interface


yebai:
## Current Interface

The existing `DifferentiateWith(f, backend)` interface in `DifferentiationInterface.jl` presents a significant limitation: it inherently supports only single-argument functions. This design makes it cumbersome to:

- Differentiate functions with multiple arguments.
- Pass additional context or non-differentiable arguments (constants, pre-allocated caches) to the differentiation backend.
    

## Proposed Interface

To address these limitations, we propose a more expressive interface for `DifferentiateWith`:

```julia
Tfunc_sig = Tuple{typeof(f), T_arg1, T_arg2, ..., T_argN}
DifferentiateWith(Tfunc_sig, backend_to_use::AbstractADType)
```

Where `Tfunc_sig` represents the function signature. The first element is the function `f` itself (or its type), and subsequent elements `T_arg1, T_arg2, ..., T_argN` represent the types of arguments to `f`.

**Argument Type Wrappers:**

To provide more context to the backend about how each argument should be treated, we can introduce wrapper types:

- **Default:** Arguments are assumed to be "active" (i.e., to be differentiated with respect to).
- `Constant{T}`: Indicates that an argument of type `T` is a constant and should not be differentiated.
- `Cache{T}`: Signals that an argument of type `T` is a pre-allocated cache that the backend can utilise.
    

**Example Usage:**

Consider a function `f(x, y, z, c)` where `x` and `y` are active arguments, `z` is a constant, and `c` is a cache. The `func_sig` would be constructed as:

```julia
Targtypes = (typeof(x), typeof(y), Constant{typeof(z)}, Cache{typeof(c)})
Tfunc_sig = Tuple{typeof(f), Targtypes...}
# or more explicitly:
# Tfunc_sig = Tuple{typeof(f), typeof(x), typeof(y), Constant{typeof(z)}, Cache{typeof(c)}}

dw = DifferentiateWith(Tfunc_sig, backend)
```

**Internal Handling:**

With this richer `Tfunc_sig`, `DifferentiateWith` can internally manage functions with multiple arguments. For backends that fundamentally operate on single-argument functions (e.g., by packing arguments into a tuple), `DifferentiateWith` can perform this packing/unpacking automatically before invoking the backend's pushforward or pullback implementations. This keeps the backend APIs simpler while providing a user-friendly multi-argument interface.

gdalle:
[...]

PR #975, opened 2026-03-04
feat: DifferentiateWith + context arguments

gdalle:
Partial fix for #806, #675.

Since DI doesn't compute derivatives with respect to contexts, in each implementation we need a way to mark the derivative as unknown.

- [x] Enrich `DifferentiateWith` with a `context_wrappers` field, denoting how additional arguments beyond `x` must be taken into account.
- [x] Implement unknown context derivatives in reverse mode:
  - With Zygote by returning a `ChainRulesCore.@not_implemented` tangent
  - With Mooncake by poisoning the `FData` and `RData` of context arguments using `NaN`
- [ ] Implement unknown context derivatives in forward mode:
  - Impossible with ForwardDiff because we don't know right away whether a context argument contains duals or not, and any `Constant` context containing `Dual`s would make the derivative wrong. We could allow `Cache`s though.
  - Possible with Mooncake but the `frule!!` doesn't yet exist for `DifferentiateWith`
- [x] Add DI tests
- [ ] Add native tests: impossible because we are missing some tangents?


gdalle:
@Technici4n I'd love to have some feedback on this, and the related https://github.com/chalk-lab/Mooncake.jl/issues/548 it enables!

Technici4n:
I'm not very familiar with contexts but the basic idea sounds reasonable. Why is the nan poisoning only used now and wasn't previously? Does this PR introduce a new kind of context?

gdalle:
The idea behind contexts is summed up concisely in https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/explanation/arguments/.
Essentially, DI only computes derivatives with respect to the first argument, but it can take into account additional information, either constants or caches that play a role in the computation but with respect to which the derivatives are not obtained. When writing a rule for a multi-argument function based on an underlying DI call, we thus have to signal that the pullback wrt the first argument is the only one that is actually correct and usable. The rest of the arguments (constants and caches) thus get `NaN`-ified pullbacks.

PR #987, opened 2026-04-02
fix `jacobian(_, AutoFiniteDiff(fdjtype=Val(:central)))`

### LINKED ISSUE #983: `FiniteDiff` with `:central` includes junk
```julia
julia> using DifferentiationInterface

julia> using FiniteDiff

julia> foo(x) = [2*x[1], 3*x[2], 4*x[1]]
foo (generic function with 1 method)

julia> jac = jacobian(foo, AutoFiniteDiff(fdjtype=Val(:central)), zeros(2))
3×2 Matrix{Float64}:
 2.0           3.94723e-309
 6.73119e-317  3.0
 4.0           7.89446e-309

julia> FiniteDiff.finite_difference_jacobian(foo, zeros(2), Val{:central}())
3×2 Matrix{Float64}:
 2.0  0.0
 0.0  3.0
 4.0  0.0
```

But `:forward` works ok:
```julia
julia> jac = jacobian(foo, AutoFiniteDiff(fdjtype=Val(:forward)), zeros(2))
3×2 Matrix{Float64}:
 2.0  0.0
 0.0  3.0
 4.0  0.0
```

oscardssmith:
fixes https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/983#issuecomment-4179203508
FiniteDiff requires `x1==x` for central mode autodiff.

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
@@ -346,9 +346,12 @@ function DI.prepare_jacobian_nokwarg(
     ) where {C}
     _sig = DI.signature(f, backend, x, contexts...; strict)
     fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
-    y = fc(x)
-    x1 = similar(x)
-    fx = similar(y)
+    x1 = if backend.fdjtype == Val(:Central)
+        copy(x)
+    else
+        similar(x)
+    end
+    fx = fc(x)
     fx1 = similar(y)
     cache = JacobianCache(x1, fx, fx1, fdjtype(backend))
     relstep = if isnothing(backend.relstep)


gdalle:
If that's true, this is not the correct fix, because preparation can be reused at different points than the original one. The correct fix would be something like `copyto!(prep.cache.x1, x)` at the beginning of execution. But this seems like something FiniteDiff should do itself, otherwise the cache is not safe to reuse in FiniteDiff either

gdalle:
Also, we would need to fix all operators and add tests

ChrisRackauckas:
The actual question is just two things, one, are you passing `f_in`? https://github.com/JuliaDiff/FiniteDiff.jl/blob/master/src/jacobians.jl#L281-L297 If you're passing `f(x)` but it's the wrong value, yeah that's a problem. If your problem is "but I don't know how to keep it up to date and ..." then... just don't pass it here and let it evaluate it? It's an interface for skipping an `f` evaluation, and if that interface doesn't apply to DI, then don't use it.

Secondly, the actual issue from the user looks like just uninitialized memory. FiniteDiff doesn't zero things in its non-allocating functions because it assumes the caller has setup the memory appropriately and doesn't touch things it doesn't differentiate. The solution is just to `J .= 0`.

oscardssmith:
@ChrisRackauckas the problem isn't the jacobian. It's `x1` which when using `:central` for differentiation is required to be equal to `x`.

gdalle:
I still don't understand why we shouldn't fix that on the FiniteDiff side?

oscardssmith:
It's a fairly big and arguably breaking change on the FiniteDiff side, but IMO it's probably the way to go. The current FiniteDiff api is very hard to use correctly here.

gdalle:
Why would it be breaking to be more careful than we currently are regarding cache updates?

gdalle:
@oscardssmith if that's okay I'm gonna close this in favor of https://github.com/JuliaDiff/FiniteDiff.jl/issues/213, since we agree that the solution should live there instead of being a DI hot patch

gdalle:
Should be fixed by https://github.com/JuliaDiff/FiniteDiff.jl/pull/214

PR #988, opened 2026-04-04
Mooncake 0.5.25 compat

sunxd3:
An attempt at addressing https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/986.

Feel free to make any edits or take over!

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
@@ -41,7 +41,8 @@ function DI.value_and_pushforward(
             map(first_unwrap, contexts, prep.context_tangents)...,
         )
         y = first(y_and_dy)
-        dy = _maybe_to_primal(last(y_and_dy), y)
+        dy_raw = last(y_and_dy)
+        dy = @something(_to_friendly_value(dy_raw, y), _copy_output(dy_raw))
         return y, dy
     end
     y = _copy_output(first(ys_and_ty[1]))
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
@@ -41,7 +41,7 @@ function DI.value_and_pushforward(
             map(first_unwrap, contexts, prep.context_tangents)...,
         )
         y = first(y_and_dy)
-        dy = _copy_output(last(y_and_dy))
+        dy = _maybe_to_primal(last(y_and_dy), y)
         return y, dy
     end
     y = _copy_output(first(ys_and_ty[1]))
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
@@ -25,6 +25,7 @@ end
 # convert it to a primal-shaped value. No-op for already-converted results.
 _maybe_to_primal(tx, x) = _copy_output(tx)
 _maybe_to_primal(tx::Mooncake.Tangent, x) = tangent_to_user_primal(tx, x)
+_maybe_to_primal(tx::Mooncake.MutableTangent, x) = tangent_to_user_primal(tx, x)
 
 @inline maybe_getfield(mod, name::Symbol) =
     isdefined(mod, name) ? getfield(mod, name) : nothing
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
@@ -21,18 +21,21 @@ function zero_tangent_or_primal(x, backend::AnyAutoMooncake)
     end
 end
 
-@inline maybe_getfield(mod, name::Symbol) = isdefined(mod, name) ? getfield(mod, name) : nothing
+@inline maybe_getfield(mod, name::Symbol) =
+    isdefined(mod, name) ? getfield(mod, name) : nothing
 
-const mooncake_tangent_to_friendly = maybe_getfield(Mooncake, Symbol("tangent_to_friendly!!"))
+const mooncake_tangent_to_friendly = maybe_getfield(
+    Mooncake, Symbol("tangent_to_friendly!!")
+)
 const mooncake_friendly_tangent_cache = maybe_getfield(Mooncake, :FriendlyTangentCache)
 const mooncake_as_primal = maybe_getfield(Mooncake, :AsPrimal)
 const mooncake_no_cache = maybe_getfield(Mooncake, :NoCache)
 
 function tangent_to_user_primal(tx, x)
     if !isnothing(mooncake_tangent_to_friendly) &&
-       !isnothing(mooncake_friendly_tangent_cache) &&
-       !isnothing(mooncake_as_primal) &&
-       !isnothing(mooncake_no_cache)
+        !isnothing(mooncake_friendly_tangent_cache) &&
+        !isnothing(mooncake_as_primal) &&
+        !isnothing(mooncake_no_cache)
         dest = mooncake_friendly_tangent_cache{mooncake_as_primal}(_copy_output(x))
         cache = isbitstype(typeof(x)) ? mooncake_no_cache() : IdDict{Any,Any}()
         return mooncake_tangent_to_friendly(dest, x, tx, cache)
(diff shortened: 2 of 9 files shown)


sunxd3:
I think I understand the CI error, there is something I need to patch on the Mooncake side, will come back to this

sunxd3:
https://github.com/chalk-lab/Mooncake.jl/pull/1129 should unblock this PR, we'll release it as soon as it's available

gdalle:
Thank you for taking a crack at this! I'll wait until the tests pass before reviewing if that's okay

sunxd3:
totally fine!

sunxd3:
@gdalle the Mooncake CIs are passing (it probably requires Mooncake v0.5.26). 

Could you take over? I also won't be offended if you want to start a new PR.

gdalle:
I'll take a look when I can, thanks a bunch! DI's tests are failing on main too because of Mooncake's breaking release so this is a priority for me. Do you know why coverage is not complete?

sunxd3:
Thanks a lot.

> Do you know why coverage is not complete?

I am not certain. A bad guess is that some code changes are more defensive than necessary. Sorry!

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
```diff
@@ -11,9 +11,41 @@ end
...
+
+@inline maybe_getfield(mod, name::Symbol) =
+    isdefined(mod, name) ? getfield(mod, name) : nothing
+
+const mooncake_tangent_to_friendly = maybe_getfield(
+    Mooncake, Symbol("tangent_to_friendly!!")
+)
+const mooncake_friendly_tangent_cache = maybe_getfield(Mooncake, :FriendlyTangentCache)
+const mooncake_as_primal = maybe_getfield(Mooncake, :AsPrimal)
+const mooncake_no_cache = maybe_getfield(Mooncake, :NoCache)
```
gdalle: This doesn't seem to be very robust? I'd rather impose a lower bound for Mooncake at v0.5.25 in `Project.toml` (that way we're sure we can use all of these symbols)

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
```diff
@@ -41,7 +41,7 @@ function DI.value_and_pushforward(
             map(first_unwrap, contexts, prep.context_tangents)...,
         )
         y = first(y_and_dy)
-        dy = _copy_output(last(y_and_dy))
+        dy = _maybe_to_primal(last(y_and_dy), y)
```
gdalle: Why do we need to ensure that primal conversion happens here? If `friendly_tangents` is set to `true`, won't Mooncake's pushforward and pullback already return a primal-like object?

DifferentiationInterface/test/Back/Mooncake/test.jl
```diff
@@ -78,5 +79,36 @@ test_differentiation(
...
     logging = LOGGING,
-    excluded = SECOND_ORDER
+    excluded = SECOND_ORDER,
 )
+
+@testset "Friendly tangents structured matrices" begin
+    backend = AutoMooncake(; config = Mooncake.Config(; friendly_tangents = true))
+    inputs = (
+        Symmetric([2.0 1.0; 1.0 3.0]),
+        Hermitian(ComplexF64[2 1 + im; 1 - im 3]),
```
gdalle: Do you know which convention Mooncake uses for gradients of functions with complex inputs and real outputs? There are two possible choices, see e.g. https://arxiv.org/abs/2409.06752

DifferentiationInterface/test/Back/Mooncake/test.jl
```diff
@@ -78,5 +79,36 @@ test_differentiation(
...
+        SymTridiagonal([2.0, 3.0, 4.0], [5.0, 6.0]),
+    )
+    f(x) = real(sum(abs2, x))
+
+    @testset "$(typeof(x))" for x in inputs
+        grad = gradient(f, backend, x)
+        y, grad2 = value_and_gradient(f, backend, x)
+        pb = only(pullback(identity, backend, x, (x,)))
```
gdalle, line 170: This is not a strong enough test, the function is too simple

DifferentiationInterface/test/Back/Mooncake/test.jl
```diff
@@ -78,5 +79,36 @@ test_differentiation(
...
+
+        @test grad isa Matrix
+        @test grad2 isa Matrix
+        @test pb isa Matrix
+        @test grad == grad2
```
gdalle, line 175: `grad` and `grad2` are never compared against the ground truth

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
```diff
@@ -11,9 +11,41 @@ end
...
+
+function tangent_to_user_primal(tx, x)
+    if !isnothing(mooncake_tangent_to_friendly) &&
+            !isnothing(mooncake_friendly_tangent_cache) &&
+            !isnothing(mooncake_as_primal) &&
+            !isnothing(mooncake_no_cache)
+        dest = mooncake_friendly_tangent_cache{mooncake_as_primal}(_copy_output(x))
+        cache = isbitstype(typeof(x)) ? mooncake_no_cache() : IdDict{Any, Any}()
```
gdalle: Why is a type-unstable dictionary needed here?
Does this make every tangent-to-primal conversion outside of bitstypes slow?

gdalle: Regardless of why, we may want to allocate this dictionary in the preparation phase

gdalle (changes requested):
Thank you for trying to fix what others broke! I added a few remarks to understand the task a bit better, I'll wait for your answers

gdalle:
@AstitvaAggarwal @Technici4n could you maybe take a look too?

DifferentiationInterface/test/Back/Mooncake/test.jl
Technici4n: See: https://github.com/chalk-lab/Mooncake.jl/pull/773

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
AstitvaAggarwal: yup this will regress performance on hot paths. This should be allocated once during the prepare phase and stored in the extras cache.

AstitvaAggarwal (commented):
also we might want to keep track of future possible tangent_types: `_maybe_to_primal(x, _) = x` will silently pass through any tangent type not yet accounted for (e.g. a future `Mooncake.SparseTangent`), making failures invisible.

gdalle:
@sunxd3 can you explain why type piracy is needed here? Sounds like there should be a StaticArraysExt in Mooncake (but the maintainer doesn't like it) or the other way around. I will not introduce piracy in a package depended upon by the entire ecosystem.

yolhan83:
would be weird but one could do a thin DI wrapper DISArray of a SArray which would just wrap a SArray everytime it needs to go into a Mooncake function, llvm should understand this is unneeded and remove it at compile time (since its all static) but then, there is the question of do we need any kind of interface on those, and I hope not otherwise its just too big of a job.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
sunxd3: Thanks for clarifying this, bumping Mooncake compat would be cleaner, I did it out of backward compat concerns.

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
sunxd3: If I remember correctly, it was for SArray compat

DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/utils.jl
gdalle: I'll take over and try to write code that is compatible with old and new Mooncake versions, thanks!

sunxd3: Thanks, Guillaume!

sunxd3:
Thanks for clarifying the policy. Yes, this was indeed motivated by StaticArray compatibility under the current convention.

The better path forward is to add a Mooncake extension to StaticArray rather than handle it here.

Unfortunately, I won't have the bandwidth to do that properly in the near future. As much as I hate to leave this PR in limbo, I think the right call is to close it for now. 
Once StaticArray and Mooncake integration is properly done, we can revisit the Mooncake-DI integration and do it cleanly. 

Thanks for the thorough and thoughtful reviews, they should serve as great reference for the future.

sunxd3:
I think it would be easier to work in a native branch than my fork

Issue #350, opened 2024-07-13
Better error messages for wrong preparation


gdalle:
When the user prepares with the wrong backend / function, you probably want to throw nice errors

gdalle:
Done with the `strict` keyword

PR #989, opened 2026-04-10
Mooncake ext: unwrap struct tangents back to primal type

ChrisRackauckas-Claude:
## Summary

`Mooncake.value_and_gradient!!` (and the pullback / pushforward analogues) returns the differential as a `Mooncake.Tangent` / `Mooncake.MutableTangent` whenever the primal is a struct-backed array such as `ComponentArray` or an `MVector`. Downstream callers — most notably `OptimizationBase`, which preallocates a `ComponentVector` buffer and passes it to `gradient!` — expect a value with the same layout as the primal and call `copyto!` / `iterate` on it, which raised

```
MethodError: no method matching iterate(::Mooncake.Tangent{@NamedTuple{data::Vector{Float64}, axes::Mooncake.NoTangent}})
```

and broke every Optimization.jl loop that used `AutoMooncake` with ComponentArrays parameters (including all SciMLSensitivity neural ODE training tutorials).

This PR converts the tangent back to the primal type at the boundary of the DI extension via `Mooncake.tangent_to_primal!!`. This is the same unwrap path SciMLSensitivity already adopted internally (SciML/SciMLSensitivity.jl@4205d49); the helpers live in `utils.jl` and are reused by the gradient, pullback, and pushforward code paths in both `onearg.jl` / `twoarg.jl` and the forward counterparts so the conversion happens consistently at one place.

### `gradient!` with mismatched buffer types

DI's `gradient!` accepts `grad` buffers whose type differs from the primal (e.g. an `MVector` buffer for an `SVector` primal — `default_scenarios()` exercises this). The in-place helper allocates the unwrap target with `_copy_output(x)` and then `copyto!`s into `grad`. When `grad` itself is immutable (`SVector`), no in-place update is possible, so `gradient!` forwards the freshly built primal-shaped value rather than the unchanged buffer. That matches what callers compare against and is the only sensible interpretation of `gradient!` for an immutable destination.

The mutable-vs-immutable check uses `ismutabletype(typeof(parent(grad)))` because `ComponentVector` is itself an immutable struct (`ismutabletype` returns `false`) but wraps a mutable `Vector` and supports `copyto!` — walking down to the array parent captures both `ComponentVector` and `SVector` correctly.

### `tangent_to_primal!!` deprecation

`tangent_to_primal!!` is a deprecated Mooncake API; the future replacement is `tangent_to_friendly!!`, but it currently returns the raw `Tangent` for `ComponentArray` (no `friendly_tangent_cache` override exists upstream) and so is not yet a viable substitute. The DI extension already uses `tangent_to_primal!!` in `zero_tangent_or_primal`, so there is no new dependency on a deprecated symbol — only a new call site. A comment in `utils.jl` notes the migration path.

### Test results

Full Mooncake backend test suite on Julia 1.11:

| | Baseline (DI 0.7.16) | This PR |
|---|---|---|
| Passed | 35,841 | **37,077** |
| Failed | 0 | 0 |
| Errored | 184 | **0** |

The 184 baseline errors were almost entirely Mooncake returning `Tangent` / `MutableTangent` for `MVector` / `SVector` / `ComponentVector` primals in `default_scenarios()` and `static_scenarios()`. They all go away with this fix, plus the new `component_scenarios()` block adds 468 ComponentArrays-specific assertions that are also passing.

## Test plan

- [x] `JULIA_DI_TEST_GROUP=Mooncake julia run_backend.jl` — 37,077 passed, 0 failed, 0 errored
- [x] Targeted ComponentArrays gradient/gradient!/pullback/pushforward via `component_scenarios()` — all pass
- [x] End-to-end SciMLSensitivity neural ODE training (Lux + ComponentArrays + `Optimization.OptimizationFunction(loss, AutoMooncake(...))`) — converges in 5 Adam steps from 169.5 → 12.9, previously raised the `MethodError` shown above

## Excluded scenario

`AutoMooncakeForward()` (without `friendly_tangents`) is excluded from the new `component_scenarios()` block because its forward-mode pushforward path has a separate, pre-existing bug at the *input* (Dual construction) side: it raises `ArgumentError: Tangent types do not match primal types` when given a `ComponentVector` `dx`, because Mooncake forward mode expects the tangent to already be a `Mooncake.Tangent` rather than a primal-shaped value. That input-side conversion is independent of the output-side fix in this PR; the friendly-tangents forward backend is exercised and passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/DifferentiationInterfaceMooncakeExt.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/DifferentiationInterfaceMooncakeExt.jl
@@ -23,6 +23,8 @@ using Mooncake:
     rdata,
     tangent_type,
     NoTangent,
+    Tangent,
+    MutableTangent,
     @is_primitive,
     zero_fcodual,
     MinimalCtx,
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl
@@ -41,7 +41,8 @@ function DI.value_and_pushforward(
             map(first_unwrap, contexts, prep.context_tangents)...,
         )
         y = first(y_and_dy)
-        dy = _copy_output(last(y_and_dy))
+        dy_raw = last(y_and_dy)
+        dy = _to_primal_alloc(y, dy_raw)
         return y, dy
     end
     y = _copy_output(first(ys_and_ty[1]))
@@ -72,7 +73,7 @@ function DI.value_and_pushforward!(
     ) where {F, C}
     DI.check_prep(f, prep, backend, x, tx, contexts...)
     y, new_ty = DI.value_and_pushforward(f, prep, backend, x, tx, contexts...)
-    foreach(copyto!, ty, new_ty)
+    foreach(_to_primal!, ty, new_ty)
     return y, ty
 end
 
--- a/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_twoarg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/forward_twoarg.jl
@@ -55,7 +55,7 @@ function DI.value_and_pushforward(
             (x, dx),
             map(first_unwrap, contexts, prep.context_tangents)...,
         )
-        return _copy_output(new_dy)
+        return _to_primal_alloc(y, new_dy)
     end
     return y, ty
 end
@@ -93,7 +93,7 @@ function DI.value_and_pushforward!(
             (x, dx),
             map(first_unwrap, contexts, prep.context_tangents)...,
         )
-        copyto!(dy, new_dy)
+        _to_primal!(dy, new_dy)
     end
     return y, ty
 end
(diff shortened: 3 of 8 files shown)


ChrisRackauckas-Claude:
Heads up that this is no longer a hard dependency for the SciML/SciMLSensitivity.jl#1419 docs migration. After SciML/ComponentArrays.jl#350 was merged into ComponentArrays main (registry PR JuliaRegistries/General#152661, releasing as v0.15.34), the migration also works with stock DI 0.7.16 if the user opts in to `Mooncake.Config(; friendly_tangents = true)`. I switched the SMS docs PR over to that pattern so it can ship as soon as CA 0.15.34 lands in the registry.

This patch is still independently useful though:

1. It makes the **default** `AutoMooncake(; config = nothing)` form work without the user having to know about `friendly_tangents`. With CA 0.15.34 alone, the default still crashes — only `friendly_tangents = true` is fixed (because Mooncake's `value_and_gradient!!` only invokes `friendly_tangent_cache` when the flag is on).

2. It also fixes the `MVector` / `SVector` cases (the 184 → 0 errors in the Mooncake test suite), which CA #350 doesn't touch — those are independent of ComponentArrays.

3. The `_can_setindex(grad)` fallback for immutable buffers (returning the freshly built primal-shaped value when `gradient!` is given an `SVector` destination) is novel and not duplicated by CA #350.

So I'd argue it's still worth merging — it just becomes a quality-of-life improvement rather than a hard blocker. Happy to rebase / split / scope down if you'd prefer a smaller change.

gdalle:
Is this the same fix @sunxd3 is working on in #988?

gdalle:
@ChrisRackauckas given the discussion in https://github.com/SciML/SciMLSensitivity.jl/pull/1419 and the existence of #988, is my review necessary on this one?

PR #1019, opened 2026-05-20
fix: make FiniteDiff prep non-allocating with static arrays

dingraha:
I noticed some allocations when using FiniteDiff with StaticArrays that I eventually traced to calls to `similar` in the FiniteDiff extension. Adding the check for immutable input or output arrays seemed to fix this. Haven't attempted to add tests for this—wanted to get some feedback before diving in. What do you think?


gdalle:
Nice catch! The alternative would be replacing `similar` with `copy`, does that work on `SArray`s  too?

dingraha:
Yes, `copy` returns a `SArray` (unlike `similar`), but that doesn't eliminate all the allocations, I think because the `FiniteDiff.JVPCache` `struct` is `mutable`. Here is a short script that shows this:

```julia
using FiniteDiff
using StaticArrays
function doit()
    backend = AutoFiniteDiff()
    x = @SVector [1.0, 2.0]
    tx = (2.0.*x,)
    f(x) = @. 3.0 * x
    # prep = DifferentiationInterface.prepare_pushforward(f, backend, x, tx);
    prep =  DifferentiationInterface.prepare_pushforward_nokwarg(Val(true), f, backend, x, tx)
    return prep
end
```

When `DI.prepare_pushforward_nokwarg` looks like this:

```julia
function DI.prepare_pushforward_nokwarg(
        strict::Val, f, backend::AutoFiniteDiff, x, tx::NTuple, contexts::Vararg{DI.Context, C}
    ) where {C}
    _sig = DI.signature(f, backend, x, tx, contexts...; strict)
    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
    y = fc(x)
    cache = if x isa Number || y isa Number
        nothing
    else
        JVPCache(copy(x), y, fdtype(backend))
    end
    relstep = if isnothing(backend.relstep)
        default_relstep(fdtype(backend), eltype(x))
    else
        backend.relstep
    end
    absstep = if isnothing(backend.absstep)
        relstep
    else
        backend.absstep
    end
    dir = backend.dir
    return FiniteDiffOneArgPushforwardPrep(_sig, cache, relstep, absstep, dir)
end
```

I see these allocations:

```julia
julia> @allocated prep = doit()
18134560

julia> @allocated prep = doit()
96

julia> 
```

(I assume the first call includes the allocations associated with compiling.)

Then, if I add those `ismutable_array` checks:

```julia
function DI.prepare_pushforward_nokwarg(
        strict::Val, f, backend::AutoFiniteDiff, x, tx::NTuple, contexts::Vararg{DI.Context, C}
    ) where {C}
    _sig = DI.signature(f, backend, x, tx, contexts...; strict)
    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
    y = fc(x)
    cache = if x isa Number || y isa Number || (! DI.ismutable_array(x)) || (! DI.ismutable_array(y))
        nothing
    else
        JVPCache(copy(x), y, fdtype(backend))
    end
    relstep = if isnothing(backend.relstep)
        default_relstep(fdtype(backend), eltype(x))
    else
        backend.relstep
    end
    absstep = if isnothing(backend.absstep)
        relstep
    else
        backend.absstep
    end
    dir = backend.dir
    return FiniteDiffOneArgPushforwardPrep(_sig, cache, relstep, absstep, dir)
end
```

I see this in the REPL:

```julia
julia> @allocated prep = doit()
751888

julia> @allocated prep = doit()
0

julia> 
```

Here's another example that isolates the `FiniteDiff.JVPCache`:

```julia
julia> function doit2()
                x = @SVector [1.0, 2.0]
                y = @SVector [2.0, 3.0]
                foo = Val(:forward)
                cache = FiniteDiff.JVPCache(x, y, foo)
                return cache
              end
doit2 (generic function with 1 method)

julia> @allocated cache = doit2()
882800

julia> @allocated cache = doit2()
48

julia> 
```

gdalle:
I tried implementing an upstream fix in https://github.com/JuliaDiff/FiniteDiff.jl/pull/216

dingraha:
@gdalle Excellent, thanks! That fixes the allocations I was seeing with this small example:

 ```julia
julia> function doit2()
                x = @SVector [1.0, 2.0]
                y = @SVector [2.0, 3.0]
                foo = Val(:forward)
                cache = FiniteDiff.JVPCache(x, y, foo)
                return cache
              end
doit2 (generic function with 1 method)

julia> @allocated cache = doit2()
882800

julia> @allocated cache = doit2()
48

julia> 
```

To avoid the allocations with this example

```julia
using FiniteDiff
using StaticArrays
function doit()
    backend = AutoFiniteDiff()
    x = @SVector [1.0, 2.0]
    tx = (2.0.*x,)
    f(x) = @. 3.0 * x
    # prep = DifferentiationInterface.prepare_pushforward(f, backend, x, tx);
    prep =  DifferentiationInterface.prepare_pushforward_nokwarg(Val(true), f, backend, x, tx)
    return prep
end
```

it looks like we'd need to replace the `similar` calls with `copy`, like you suggested. I updated the PR to reflect this.

Interestingly, it looks like `FiniteDiff.finite_difference_jvp` doesn't actually use the `cache::JVPCache` argument at all, so I guess it doesn't matter what's in there.

gdalle:
Thanks!
Two questions:

- Any idea why you need the internal form of `prepare_pushforward(Val(true), ...)` to avoid allocations? Is it because passing `strict` as a keyword argument causes some form of dynamic dispatch?
- Can you add a small test in the FiniteDiff test file to make sure that we have no regression?

gdalle:
@dingraha is this ready for another review?

dingraha:
@gdalle Yes, sorry for the delay.

DifferentiationInterface/test/Back/FiniteDiff/allocations.jl
```diff
@@ -0,0 +1,39 @@
...
+    function jacobian_allocs()
+        backend = AutoFiniteDiff()
+        x = @SVector [1.0, 2.0]
+        f(x) = 3.0 .* x
+        prep = DifferentiationInterface.prepare_jacobian(f, backend, x)
+        return prep
+    end
+    jacobian_allocs()
+    allocs = @allocated prep = jacobian_allocs()
+    # Using FiniteDiff.jl with StaticArrays to calculate a Jacobian does result in some allocations, apparently because the `FiniteDiff.JacobianCache` is a `mutable struct`.
```
gdalle, line 37: Could it also be due to https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/5572e56a5d65a9c8835c5c03fac4d0509839363a/DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl#L350-L352 ?

gdalle (approved):

DifferentiationInterface/test/Back/FiniteDiff/allocations.jl
dingraha: IIUC this PR changes those to `copy` and I still see a few allocations.

dingraha:
I think the test failures are unrelated to this PR?

DifferentiationInterface/Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DifferentiationInterface"
 uuid = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
-version = "0.7.21"
+version = "0.7.22"
```
```suggestion
version = "0.7.21"
```
v0.7.21 isn't released yet

gdalle (changes requested):

gdalle (approved):

PR #1022, opened 2026-05-21
fix(FiniteDiff): use fdhtype for hvp to match hessian accuracy

### LINKED ISSUE #1012: Bug(FiniteDiff): hvp accuracy under AutoFiniteDiff() defaults disagrees with hessian
## Description

`hvp(f, AutoFiniteDiff(), x, v)` returns values that differ substantially from the analytical Hessian-vector product, while `hessian(f, AutoFiniteDiff(), x) * v` for the same backend matches the analytical result. The two computations of `H*v` therefore disagree for the same backend.

Root cause: `AutoFiniteDiff` defaults to `fdtype = Val(:forward)`. The HVP path is the generic JVP-of-gradient fallback, which becomes a forward-FD over forward-FD composition and loses precision. The `hessian` path uses the separate `fdhtype` field (default `Val(:hcentral)`) and remains accurate.

## MWE

```julia
using FiniteDiff: FiniteDiff
using ADTypes: AutoFiniteDiff

f(x) = sum(x .^ 2)
x = [1.0, 2.0, 3.0]
v = [1.0, 0.0, 0.0]

backend = AutoFiniteDiff()

H = hessian(f, backend, x)        # [2 0 0; 0 2 0; 0 0 2]
Hv = H * v                         # [2.0, 0.0, 0.0]

hv = hvp(f, backend, x, (v,))[1]   # [5.999999..., 0.0, 0.0]
```

## Expected Behavior

`hvp(f, AutoFiniteDiff(), x, (v,))` and `hessian(f, AutoFiniteDiff(), x) * v` should produce values of comparable accuracy for the same backend and inputs.

## Actual Behavior

| `f`, with `x=[1,2,3]`, `v=[1,0,0]` | analytical `H*v` | `hessian(f, b, x)*v` | `hvp(f, b, x, (v,))` |
|---|---|---|---|
| `sum(x.^2)` | `[2, 0, 0]` | `[2.0, 0, 0]` | `[5.9999..., 0, 0]` |
| `sum(x.^3)` | `[6, 0, 0]` | `[6.0, 0, 0]` | `[-2.999..., 0, 0]` |
| `sum(x.^4)` | `[12, 0, 0]` | `[12.0, 0, 0]` | `[-3.999..., 0, 0]` |
| `x'*[1 2;3 4]*x`, `x=[1,2]` | `[2, 5]` | `[2.0, 5.0]` | `[4.0, 8.0]` |

The behavior is consistent across `hvp`, `hvp!`, `gradient_and_hvp`, `gradient_and_hvp!`, with and without `prepare_hvp`.

## Workaround

Setting `fdtype = Val(:central)` explicitly produces results that match the analytical answer to the precision expected of central FD:

```julia
backend = AutoFiniteDiff(; fdtype = Val(:central))
hvp(f, backend, x, (v,))[1]   # [1.999999..., 0, ~1e-6]
```

## Native Backend Comparison

FiniteDiff has no native HVP. The corresponding native composition (forward-FD gradient then forward-FD JVP) reproduces the DI result:

```julia
grad_fwd(x) = FiniteDiff.finite_difference_gradient(f, x, Val(:forward))
y = grad_fwd(x)
cache = FiniteDiff.JVPCache(similar(x), y, Val(:forward))
FiniteDiff.finite_difference_jvp(grad_fwd, x, v, cache)
# → [5.999999..., 0, 0]   matches DI

FiniteDiff.finite_difference_hessian(f, x) * v
# → [2.0, 0.0, 0.0]
```

So the precision loss originates in the forward-over-forward FD composition, not in the DI wrapping. The DI-level observation is that `AutoFiniteDiff`'s default `fdtype` is used for the HVP composition while `fdhtype` is used for `hessian`, producing different accuracy from the same backend instance.

For reference, `DifferentiationInterface/test/Back/FiniteDiff/test.jl` excludes `:hvp` from `test_differentiation` for `AutoFiniteDiff()` and tests HVP only via `SecondOrder(AutoFiniteDiff(; relstep = 1.0e-5), AutoFiniteDiff())` with `rtol = 1.0e-2`. `backends.md` marks `hvp` as ❌ for `AutoFiniteDiff` (no custom implementation), but does not describe the accuracy difference between `hvp` and `hessian` under the default settings.

## Possible Resolutions

- Use `fdhtype` (or `:central`) for the internal gradient/JVP when computing HVP under `AutoFiniteDiff`, so that `hvp` and `hessian` use comparable stencils.
- Add a note in `backends.md` recommending `AutoFiniteDiff(; fdtype = Val(:central))` (or `SecondOrder(AutoFiniteDiff(; relstep = 1.0e-5), AutoFiniteDiff())`) when using `hvp` with `AutoFiniteDiff`.

## Backend

- Backend: `AutoFiniteDiff()` (default `fdtype = Val(:forward)`)
- Affected operators: `hvp`, `hvp!`, `gradient_and_hvp`, `gradient_and_hvp!`
- `hessian` with same backend: matches analytical answer
- With `fdtype = Val(:central)`: HVP matches analytical answer

## Environment

- Julia 1.12.6
- DifferentiationInterface v0.7.18
- FiniteDiff v2.31.0
- ADTypes v1.22.0

---
🤖 I am a robot. This is an experiment in agentic bug-catching under the supervision of @adrhill and @gdalle ([#1008](https://github.com/gdalle/DifferentiationInterface.jl/issues/1008)). Contents may be hallucinated.

bdrhill:
## Summary

- Implements custom `hvp`, `hvp!`, `gradient_and_hvp`, and `gradient_and_hvp!` for `AutoFiniteDiff`
- Uses `fdhtype` (default `Val(:hcentral)`) for the Hessian computation instead of the generic JVP-of-gradient fallback which used `fdtype` (default `Val(:forward)`)
- The forward-over-forward finite difference composition caused significant precision loss

Fixes #1012

## Test plan

- [ ] Verify with the MWE from #1012:
  ```julia
  using DifferentiationInterface, FiniteDiff, ADTypes
  
  f(x) = sum(x .^ 2)
  x = [1.0, 2.0, 3.0]
  v = [1.0, 0.0, 0.0]
  backend = AutoFiniteDiff()
  
  H = hessian(f, backend, x)
  hv = hvp(f, backend, x, (v,))[1]
  @assert isapprox(hv, H * v, rtol=1e-6)  # should pass now
  ```
- [ ] Run FiniteDiff backend tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
@@ -550,3 +550,140 @@ function DI.value_gradient_and_hessian!(
     )
     return fc(x), grad, hess
 end
+
+## HVP
+
+struct FiniteDiffHVPPrep{SIG, C1, C2, RG, AG, RH, AH, H} <: DI.HVPPrep{SIG}
+    _sig::Val{SIG}
+    gradient_cache::C1
+    hessian_cache::C2
+    relstep_g::RG
+    absstep_g::AG
+    relstep_h::RH
+    absstep_h::AH
+    hess::H
+end
+
+function DI.prepare_hvp_nokwarg(
+        strict::Val, f, backend::AutoFiniteDiff, x, tx::NTuple, contexts::Vararg{DI.Context, C}
+    ) where {C}
+    _sig = DI.signature(f, backend, x, tx, contexts...; strict)
+    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
+    y = fc(x)
+    df = zero(y) .* x
+    gradient_cache = GradientCache(df, x, fdtype(backend))
+    hessian_cache = HessianCache(x, fdhtype(backend))
+    relstep_g = if isnothing(backend.relstep)
+        default_relstep(fdtype(backend), eltype(x))
+    else
+        backend.relstep
+    end
+    relstep_h = if isnothing(backend.relstep)
+        default_relstep(fdhtype(backend), eltype(x))
+    else
+        backend.relstep
+    end
+    absstep_g = if isnothing(backend.absstep)
+        relstep_g
+    else
+        backend.absstep
+    end
+    absstep_h = if isnothing(backend.absstep)
+        relstep_h
+    else
+        backend.absstep
+    end
+    hess = similar(x, eltype(x), (length(x), length(x)))
+    return FiniteDiffHVPPrep(
+        _sig, gradient_cache, hessian_cache, relstep_g, absstep_g, relstep_h, absstep_h, hess
+    )
+end
+
+function DI.hvp(
+        f,
+        prep::FiniteDiffHVPPrep,
+        backend::AutoFiniteDiff,
+        x,
+        tx::NTuple,
+        contexts::Vararg{DI.Context, C},
+    ) where {C}
+    DI.check_prep(f, prep, backend, x, tx, contexts...)
+    (; relstep_h, absstep_h, hess) = prep
+    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
+    finite_difference_hessian!(
+        hess, fc, x, prep.hessian_cache; relstep = relstep_h, absstep = absstep_h
+    )
+    tg = map(tx) do dx
+        hess * dx
+    end
+    return tg
+end
+
+function DI.hvp!(
+        f,
+        tg::NTuple,
+        prep::FiniteDiffHVPPrep,
+        backend::AutoFiniteDiff,
+        x,
+        tx::NTuple,
+        contexts::Vararg{DI.Context, C},
+    ) where {C}
+    DI.check_prep(f, prep, backend, x, tx, contexts...)
+    (; relstep_h, absstep_h, hess) = prep
+    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
+    finite_difference_hessian!(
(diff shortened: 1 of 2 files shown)


DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -550,3 +550,140 @@ function DI.value_gradient_and_hessian!(
...
+        prep::FiniteDiffHVPPrep,
+        backend::AutoFiniteDiff,
+        x,
+        tx::NTuple,
+        contexts::Vararg{DI.Context, C},
+    ) where {C}
+    DI.check_prep(f, prep, backend, x, tx, contexts...)
+    (; relstep_h, absstep_h, hess) = prep
+    fc = DI.fix_tail(f, map(DI.unwrap, contexts)...)
+    finite_difference_hessian!(
```
gdalle, line 613: This is much more expensive than it needs to be (order $O(n^2)$ instead of $O(n)$). The right approach is a pushforward of a gradient

gdalle (changes requested):

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
gdalle: In which case it isn't obvious which precision to choose, that of the pushforward or that of the gradient. I'm gonna close this for now

PR #1036, opened 2026-06-30
feat: Add reverse/forward counterparts

### LINKED ISSUE #1025: Backend counterparts
Implement a small utility to get the forward- or reverse-mode counterpart of a given backend

See https://github.com/rsenne/ParallelMCMC.jl/pull/32#discussion_r3323243468

rsenne:
PR to add the forward/backward counterparts discussed. I wasn't sure if for a generic backend if there was no reasonable counterpart to return e.g., if one did `reverse_counterpart(AutoForwardDiff)` if the code should throw an ArgumentError or if it should just return the backend. I opted for the former

Resolves #1025.

### DIFF
--- a/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
@@ -1,11 +1,26 @@
 ## Backend counterparts
 
-# Pin the mode while preserving the function annotation type `A`.
+# Flip the Enzyme mode while preserving its other type parameters (and the function
+# annotation `A`).
 
-function DI.forward_counterpart(::AutoEnzyme{M, A}) where {M, A}
-    return AutoEnzyme(; mode = Forward, function_annotation = A)
+function DI.forward_counterpart(
+        ::AutoEnzyme{
+            <:ReverseMode{
+                ReturnPrimal, RuntimeActivity, StrongZero, ABI, Holomorphic, ErrIfFuncWritten,
+            },
+            A,
+        },
+    ) where {ReturnPrimal, RuntimeActivity, StrongZero, ABI, Holomorphic, ErrIfFuncWritten, A}
+    mode = ForwardMode{ReturnPrimal, ABI, ErrIfFuncWritten, RuntimeActivity, StrongZero}()
+    return AutoEnzyme(; mode, function_annotation = A)
 end
 
-function DI.reverse_counterpart(::AutoEnzyme{M, A}) where {M, A}
-    return AutoEnzyme(; mode = Reverse, function_annotation = A)
+function DI.reverse_counterpart(
+        ::AutoEnzyme{
+            <:ForwardMode{ReturnPrimal, ABI, ErrIfFuncWritten, RuntimeActivity, StrongZero},
+            A,
+        },
+    ) where {ReturnPrimal, ABI, ErrIfFuncWritten, RuntimeActivity, StrongZero, A}
+    mode = ReverseMode{ReturnPrimal, RuntimeActivity, StrongZero, ABI, false, ErrIfFuncWritten}()
+    return AutoEnzyme(; mode, function_annotation = A)
 end
--- a/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
@@ -15,6 +15,22 @@ function DI.forward_counterpart(
     return AutoEnzyme(; mode, function_annotation = A)
 end
 
+function DI.forward_counterpart(
+        ::AutoEnzyme{
+            <:ReverseModeSplit{
+                ReturnPrimal, ReturnShadow, RuntimeActivity, StrongZero, Width,
+                ModifiedBetween, ABI, Holomorphic, ErrIfFuncWritten, ShadowInit,
+            },
+            A,
+        },
+    ) where {
+        ReturnPrimal, ReturnShadow, RuntimeActivity, StrongZero, Width,
+        ModifiedBetween, ABI, Holomorphic, ErrIfFuncWritten, ShadowInit, A,
+    }
+    mode = ForwardMode{ReturnPrimal, ABI, ErrIfFuncWritten, RuntimeActivity, StrongZero}()
+    return AutoEnzyme(; mode, function_annotation = A)
+end
+
 function DI.reverse_counterpart(
         ::AutoEnzyme{
             <:ForwardMode{ReturnPrimal, ABI, ErrIfFuncWritten, RuntimeActivity, StrongZero},
--- a/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
+++ b/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
@@ -1,0 +1,11 @@
+## Backend counterparts
+
+# Pin the mode while preserving the function annotation type `A`.
+
+function DI.forward_counterpart(::AutoEnzyme{M, A}) where {M, A}
+    return AutoEnzyme(; mode = Forward, function_annotation = A)
+end
+
+function DI.reverse_counterpart(::AutoEnzyme{M, A}) where {M, A}
+    return AutoEnzyme(; mode = Reverse, function_annotation = A)
+end
--- a/DifferentiationInterface/src/utils/counterparts.jl
+++ b/DifferentiationInterface/src/utils/counterparts.jl
@@ -1,26 +1,36 @@
 """
     forward_counterpart(backend)
 
-Return the forward-mode counterpart of `backend`, if it exists.
+Return a forward-mode counterpart of `backend`.
+
+If `backend` has a dedicated forward-mode counterpart (e.g. `AutoMooncake` has
+`AutoMooncakeForward`), it is returned. Else, a warning is emitted in the case the `backend`
+is reverse-mode only.
 """
 function forward_counterpart(backend::AbstractADType)
-    mode(backend) isa ReverseMode &&
-        throw(ArgumentError("No forward-mode counterpart known for `$backend`."))
+    if !(mode(backend) isa Union{ForwardMode, ForwardOrReverseMode, SymbolicMode})
+        @warn "No forward-mode counterpart known for `$backend`, returning it unchanged." maxlog =
+            1
+    end
     return backend
 end
-forward_counterpart(backend::AutoMooncake) = AutoMooncakeForward(; config = backend.config)
 
 """
     reverse_counterpart(backend)
 
-Return the reverse-mode counterpart of `backend`, if it exists.
+Return a reverse-mode counterpart of `backend`.
+
+If `backend` has a dedicated reverse-mode counterpart (e.g. `AutoMooncakeForward` has
+`AutoMooncake`), it is returned. Else, a warning is emitted in the case the `backend` is
+forward-mode only.
 """
 function reverse_counterpart(backend::AbstractADType)
-    mode(backend) isa ForwardMode &&
-        throw(ArgumentError("No reverse-mode counterpart known for `$backend`."))
+    if !(mode(backend) isa Union{ReverseMode, ForwardOrReverseMode, SymbolicMode})
+        @warn "No reverse-mode counterpart known for `$backend`, returning it unchanged." maxlog =
+            1
+    end
     return backend
 end
-reverse_counterpart(backend::AutoMooncakeForward) = AutoMooncake(; config = backend.config)
 
-# AutoEnzyme counterparts need the `Forward`/`Reverse` mode objects from EnzymeCore,
-# so they live in DifferentiationInterfaceEnzymeExt.
+# Backend-specific counterparts (e.g., for AutoMooncake and AutoEnzyme) are defined in the
+# corresponding package extensions.
--- a/DifferentiationInterface/src/utils/counterparts.jl
+++ b/DifferentiationInterface/src/utils/counterparts.jl
@@ -4,12 +4,13 @@
 Return a forward-mode counterpart of `backend`.
 
 If `backend` has a dedicated forward-mode counterpart (e.g. `AutoMooncake` has
-`AutoMooncakeForward`), it is returned. Else, a warning is emitted in the case the `backend`
-is reverse-mode only.
+`AutoMooncakeForward`), it is returned. Else, `backend` itself is returned (DI allows
+reverse-mode backends to execute pushforwards), with a warning in the case it is
+reverse-mode only.
 """
 function forward_counterpart(backend::AbstractADType)
     if !(mode(backend) isa Union{ForwardMode, ForwardOrReverseMode, SymbolicMode})
-        @warn "No forward-mode counterpart known for `$backend`, returning it unchanged." maxlog =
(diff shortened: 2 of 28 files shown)


DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
```diff
@@ -0,0 +1,11 @@
+## Backend counterparts
+
+# Pin the mode while preserving the function annotation type `A`.
```
gdalle: Actually for Enzyme we have to be a bit careful and copy the mode faithfully: if you replace `M` with just `EnzymeCore.Forward`, you lose all the other attributes of that mode. See the definitions in [this file](https://github.com/EnzymeAD/Enzyme.jl/blob/8de06ee1466cc060818c9b65b5a9760178991413/lib/EnzymeCore/src/EnzymeCore.jl) to understand better. Typically, what we need here is something like
```julia
function DI.forward_counterpart(
    ::AutoEnzyme{
        <:ReverseMode{
            ReturnPrimal,RuntimeActivity,StrongZero,ABI,Holomorphic,ErrIfFuncWritten
        },
        A,
    },
) where {ReturnPrimal,RuntimeActivity,StrongZero,ABI,Holomorphic,ErrIfFuncWritten,A}
    return AutoEnzyme(;
        mode=ForwardMode{ReturnPrimal,ABI,ErrIfFuncWritten,RuntimeActivity,StrongZero},
        function_annotation=A,
    )
end
```
It relies on Enzyme internals so it's not ideal (see https://github.com/EnzymeAD/Enzyme.jl/issues/1881), but the fact that I put `<:ReverseMode` in the signature should at least ensure that _additional_ type parameters won't suddenly break DI.

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    forward_counterpart(backend)
+
+Return the forward-mode counterpart of `backend`, if it exists.
+"""
+function forward_counterpart(backend::AbstractADType)
+    mode(backend) isa ReverseMode &&
+        throw(ArgumentError("No forward-mode counterpart known for `$backend`."))
```
gdalle: I'd rather return the backend itself (possibly with a `@warn`?) since DI allows forward-mode backends to execute pullbacks and vice-versa

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    forward_counterpart(backend)
+
+Return the forward-mode counterpart of `backend`, if it exists.
+"""
+function forward_counterpart(backend::AbstractADType)
+    mode(backend) isa ReverseMode &&
```
gdalle: You're forgetting `ForwardOrReverseMode` and `SymbolicMode`

DifferentiationInterface/test/Core/Internals/backends.jl
```diff
@@ -45,6 +47,15 @@ end
     @test_throws MethodError pullback_performance(backend)
 end
 
+@testset "Counterparts" begin
```
gdalle, line 58: We should also test this for every backend in the corresponding test, same as I do for `is_available`. Might be time to automate that with a small script in DIT

DifferentiationInterface/test/Back/Mooncake/test.jl
```diff
@@ -21,6 +21,13 @@ for backend in backends
     @test check_inplace(backend)
 end
 
+@testset "Counterparts" begin
+    config = Mooncake.Config(; friendly_tangents = true)
+    @test DifferentiationInterface.forward_counterpart(AutoMooncake()) === AutoMooncakeForward()
+    @test DifferentiationInterface.forward_counterpart(AutoMooncake(; config)) ===
```
gdalle, line 30: I think you forgot to write this in the Mooncake extension?
Also, we want to test that we carry over the `config`

DifferentiationInterface/test/Back/Enzyme/test.jl
```diff
@@ -30,6 +30,14 @@ duplicated_backends = [
     end
 end;
 
+@testset "Counterparts" begin
+    rev = AutoEnzyme(; mode = Enzyme.Reverse, function_annotation = Enzyme.Const)
+    fwd = DifferentiationInterface.forward_counterpart(rev)
+    @test ADTypes.mode(fwd) isa ADTypes.ForwardMode
+    @test fwd isa AutoEnzyme{<:Any, Enzyme.Const}  # function annotation preserved
```
gdalle, line 40: We also want to test that mode properties are preserved (see long remark above)

gdalle (changes requested):
Thank you for getting this started!
Most test failures are a red herring (test CI runs by forcing the highest possible version _of each package_, which here leads to incompatibilities between the recent OrderedCollections v2.0 and the not-yet-updated DataFrames)

rsenne:
I think I addressed everything:

- We now emit a warning (and return the backend unchanged) instead of throwing an error
- Added a testing recipe (test_counterparts) in DIT that I now call in each backend's tests
- Moved the `Mooncake` counterparts into the extension
- Tested that the `Mooncake` config is carried over to the counterpart
- Tested that the Enzyme mode attributes are preserved across the flip, plus the `<:ReverseMode` signature so extra type params won't break DI
- Generic fallbacks now also cover `ForwardOrReverseMode` and `SymbolicMode`

Let me know if anything needs further adjustment!

DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
```diff
@@ -0,0 +1,26 @@
+## Backend counterparts
+
+# Flip the Enzyme mode while preserving its other type parameters (and the function
```
gdalle, line 3: We also need to handle [`ReverseModeSplit`](https://github.com/EnzymeAD/Enzyme.jl/blob/65063ccd1a20ea0303caac42634b3a9db41b9635/lib/EnzymeCore/src/EnzymeCore.jl#L464) and map it to `ForwardMode`

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,36 @@
...
+Return a forward-mode counterpart of `backend`.
+
+If `backend` has a dedicated forward-mode counterpart (e.g. `AutoMooncake` has
+`AutoMooncakeForward`), it is returned. Else, a warning is emitted in the case the `backend`
+is reverse-mode only.
+"""
+function forward_counterpart(backend::AbstractADType)
+    if !(mode(backend) isa Union{ForwardMode, ForwardOrReverseMode, SymbolicMode})
+        @warn "No forward-mode counterpart known for `$backend`, returning it unchanged." maxlog =
+            1
```
gdalle:
```suggestion
        @warn "The forward-mode counterpart of `$backend` is itself, returning it unchanged." maxlog = 1
```

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,36 @@
...
+Return a reverse-mode counterpart of `backend`.
+
+If `backend` has a dedicated reverse-mode counterpart (e.g. `AutoMooncakeForward` has
+`AutoMooncake`), it is returned. Else, a warning is emitted in the case the `backend` is
+forward-mode only.
+"""
+function reverse_counterpart(backend::AbstractADType)
+    if !(mode(backend) isa Union{ReverseMode, ForwardOrReverseMode, SymbolicMode})
+        @warn "No reverse-mode counterpart known for `$backend`, returning it unchanged." maxlog =
+            1
```
gdalle:
```suggestion
        @warn "The reverse-mode counterpart of `$backend` is itself, returning it unchanged." maxlog = 1
```

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,36 @@
...
+    end
+    return backend
+end
+
+# Backend-specific counterparts (e.g., for AutoMooncake and AutoEnzyme) are defined in the
+# corresponding package extensions.
```
gdalle: There are a few more counterparts we need to implement and test in DI core:

- `ADTypes.NoAutoDiff` => handled correctly by the defaults above
- `ADTypes.AutoSparse` => grab `dense_backend`, apply counterpart and rewrap?
- `DI.SecondOrder` => no idea, cause applying the counterpart to both `inner` and `outer` doesn't mean much. Maybe best to error here
- `DI.MixedMode` => do nothing
- `DI.AutoForwardFromPrimitive` => `DI.AutoReverseFromPrimitive` (and apply counterpart inside? not sure)
- `DI.AutoReverseFromPrimitive` =>`DI.AutoForwardFromPrimitive` (and apply counterpart inside? not sure)

DifferentiationInterface/test/Core/Internals/backends.jl
```diff
@@ -45,6 +47,15 @@ end
...
+    # forward-/reverse-mode backends are their own counterpart
+    @test forward_counterpart(fb) === fb
+    @test reverse_counterpart(rb) === rb
+    # without a known counterpart, the backend is returned unchanged (with a warning)
```
gdalle: Should we test that the warning is emitted? I'm honestly not sure it's even necessary to emit it in the first place

DifferentiationInterfaceTest/src/tests/counterparts.jl
```diff
@@ -0,0 +1,18 @@
...
+unchanged (idempotence on the counterpart's own output).
+"""
+function test_counterparts(backend::AbstractADType)
+    @testset "Counterparts: $(typeof(backend))" begin
+        fc = forward_counterpart(backend)
+        rc = reverse_counterpart(backend)
+        @test check_available(fc)
+        @test check_available(rc)
+        @test mode(forward_counterpart(fc)) == mode(fc)
+        @test mode(reverse_counterpart(rc)) == mode(rc)
```
gdalle, line 30: Technically you're not testing that `forward_counterpart` returns a forward-mode backend here: just that applying it a second time doesn't change the mode

DifferentiationInterfaceTest/src/DifferentiationInterfaceTest.jl
```diff
@@ -137,6 +140,7 @@ include("scenarios/empty.jl")
 include("scenarios/extensions.jl")
 
 include("tests/correctness_eval.jl")
+include("tests/counterparts.jl")
```
gdalle, line 143: I think we might have an issue with code coverage if we don't test `test_counterparts` in DIT's test suite

gdalle (changes requested):
Almost there!

DifferentiationInterface/test/Core/Internals/backends.jl
rsenne: I added a test--but re should we emit a warning, I'm not completely sure. On one hand I do think the behavior would be a bit unintuitive, but I also think it could get annoying if as you stated, its intentional for reasons of pullbacks etc. Another approach is put an admonition in the docstring perhaps?

rsenne:
okay think i got everything again

DifferentiationInterface/src/utils/counterparts.jl
rsenne: I basically went with your judgement on everything here as I don;t know the functionality enough to have strong opinions. That said -- your suggestions all made sense to me FWIW

DifferentiationInterface/test/Core/Internals/backends.jl
gdalle: Yeah let's get rid of the warning and just admonish in the docstring

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,85 @@
...
+Return a forward-mode counterpart of `backend`.
+
+If `backend` has a dedicated forward-mode counterpart (e.g. `AutoMooncake` has
+`AutoMooncakeForward`), it is returned. Else, `backend` itself is returned (DI allows
+reverse-mode backends to execute pushforwards), with a warning in the case it is
+reverse-mode only.
+"""
+function forward_counterpart(backend::AbstractADType)
+    if !(mode(backend) isa Union{ForwardMode, ForwardOrReverseMode, SymbolicMode})
+        @warn "The forward-mode counterpart of `$backend` is itself, returning it unchanged." maxlog =
```
gdalle: You're right, warnings will get annoying quickly, let's get rid of them

DifferentiationInterface/src/utils/counterparts.jl
```diff
@@ -0,0 +1,85 @@
...
+Return a reverse-mode counterpart of `backend`.
+
+If `backend` has a dedicated reverse-mode counterpart (e.g. `AutoMooncakeForward` has
+`AutoMooncake`), it is returned. Else, `backend` itself is returned (DI allows
+forward-mode backends to execute pullbacks), with a warning in the case it is
+forward-mode only.
+"""
+function reverse_counterpart(backend::AbstractADType)
+    if !(mode(backend) isa Union{ReverseMode, ForwardOrReverseMode, SymbolicMode})
+        @warn "The reverse-mode counterpart of `$backend` is itself, returning it unchanged." maxlog =
```
gdalle: Same as above

DifferentiationInterfaceTest/test/zero_backends.jl
```diff
@@ -7,6 +7,11 @@ using DifferentiationInterfaceTest
 using DifferentiationInterfaceTest: allocfree_scenarios, no_matrices
 using Test
 
+## Counterparts
+
+test_counterparts(AutoZeroForward())
+test_counterparts(AutoZeroReverse())
```
gdalle, line 13: Why doesn't this show up in code coverage?

gdalle (changes requested):
LGTM except
- the warnings to remove
- the test failures for Zygote and ReverseDiff
- the missing coverage

DifferentiationInterfaceTest/test/zero_backends.jl
rsenne: No idea--have noticed CodeCov acting up lately, unclear why

gdalle (approved):
Thanks, sorry for the long-winded review!

DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/counterparts.jl
```diff
@@ -0,0 +1,42 @@
...
+# annotation `A`).
+
+function DI.forward_counterpart(
+        ::AutoEnzyme{
+            <:ReverseMode{
+                ReturnPrimal, RuntimeActivity, StrongZero, ABI, Holomorphic, ErrIfFuncWritten,
```
wsmoses, line 9: this unquestionably uses internal apis and is subject to breakage.

@gdalle / @rsenne instead can I recommend you add a version of this in enzymecore [and then call that version here]. That way it won't get broken if/when we add/change type parameters to the mode

rsenne: I can author an upstream PR -- that work for you @gdalle?

gdalle: Yeah I have already mentioned it above in the PR, but DI uses these internals at other points cause it definitely needs them. We'll upstream all we can to EnzymeCore, feel free to get the ball rolling @rsenne and I'll follow up

PR #1050, opened 2026-07-29
Fix mangled contributors list

rsenne:
I was looking at the docs and noticed the Contributors was not rendering properly. When the README is converted to the `index.md` file it wasn't respecting the HTML so I wrapped in a raw HTML block. Also, when i clicked on the badge to quickly see if it was fixed i noticed that didn't work (not that thats a huge deal) so i then realized that there was an issue in how github vs. documentor handled the heading anchors. both are now fixed and i checked the docs local


rsenne:
image

as it appears locally now!

gdalle (approved):
thanks!

PR #1053, opened 2026-08-08
update HyperHessians extension for 0.3

KristofferC:
Maybe requires https://github.com/SciML/ADTypes.jl/pull/176 to merge + tag first?

- forward the new simd and jet backend options into HessianConfig /
  HVPConfig (jet selects the symmetric Jet representation for Hessians;
  HVPs stay chunked HyperDuals)
- HyperDual gained a simd backend type parameter; use the scalar
  default HyperDual{1, 1, T, false} for cache translation
- preserve simd/jet when threshold_batchsize rebuilds the backend
- bump HyperHessians compat to 0.3 and test the new option combinations


DifferentiationInterface/Project.toml
```suggestion
ADTypes = "1.23.0"
```

gdalle (approved):
Thanks!

PR #1054, opened 2026-08-10
fix: avoid duplicate function calls with FiniteDiff

### LINKED ISSUE #4162: AutoFiniteDiff JVP recomputes f(x) on every Krylov matvec; nf accounting double-counts it
## Summary

On the matrix-free Krylov path with `autodiff = AutoFiniteDiff()`, the base evaluation `f(x)` is recomputed on **every** Krylov matvec, even though `x` is fixed for the entire linear solve. `FiniteDiff.finite_difference_jvp!` accepts an `f_in` argument precisely so the caller can supply a cached `f(x)`, but DifferentiationInterface's `pushforward!` ([`DifferentiationInterfaceFiniteDiffExt/twoarg.jl:113-131`](https://github.com/JuliaDiff/DifferentiationInterface.jl)) never passes it.

## Measured cost

2D Brusselator, N=32 (2048 ODEs), `FBDF(linsolve = KrylovJL_GMRES())`, `reltol = abstol = 1e-6`, Julia 1.12.4, `BLAS.set_num_threads(1)`:

| | time/matvec | raw `f` calls per matvec |
|---|---|---|
| plain `f` | 9.79 µs | — |
| `AutoForwardDiff` JVP | 27.67 µs | 1 (dual) |
| `AutoFiniteDiff` JVP | 24.96 µs | 2 |

One of those two `f` calls is the redundant base evaluation. In an end-to-end solve with a correctly scaled JVP step, caching it moved wall time from **0.97× to 0.83×** relative to `AutoForwardDiff` — about a 1.2× improvement — with bit-identical step statistics.

It also halves the reported `nf`. `dolinsolve` accounts `2 * linres.iters` for the finite-difference path versus `1 * linres.iters` for AD, so `nf` currently overstates the real cost difference by 2× and makes `AutoFiniteDiff` look worse in solver statistics than it is.

## Context

Found while root-causing a separate and much larger problem on the same path: `FiniteDiff`'s JVP step size is dimensionally wrong (`sqrt(|x⋅v|)` has units `√([x][v])`, not `[x]/[v]`), which pins the step at the `absstep` floor regardless of state scale and drives Newton convergence failures worth several hundred× in wall time. That is being fixed separately in FiniteDiff.jl. **This issue is the smaller, independent inefficiency** — worth ~1.2× on its own, and worth fixing regardless of the step-size work.

## Where the fix belongs

Primarily in DifferentiationInterface's FiniteDiff extension: thread a cached base evaluation through the prep object and pass it as `f_in`. That is a different repo; filing here because this is where the cost lands and where the `nf` accounting lives, and because the `nf` double-count in `dolinsolve` is fixable here independently.

For reference, the pre-DifferentiationInterface implementation handled this: `SparseDiffTools.num_jacvec!` carried a `compute_f0` flag to skip the redundant evaluation. Both that and the correct step scaling were lost in the migration (commit `13c036b2f4`).

## Environment

```
Julia 1.12.4
OrdinaryDiffEq v7.3.0
OrdinaryDiffEqDifferentiation 3.7.0
DifferentiationInterface 0.7.20
FiniteDiff 2.32.1
```

`prepare_jvp` and `dolinsolve` on master are byte-identical to the released versions, so this applies to master unchanged.

gdalle:
Meant to fix https://github.com/SciML/OrdinaryDiffEq.jl/issues/4162 by making one function call inside `DI.prepare_pushforward_samepoint` and then reusing it inside `FiniteDiff.finite_difference_jvp`


DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
```diff
@@ -30,7 +32,32 @@ function DI.prepare_pushforward_nokwarg(
...
+        f,
+        prep::FiniteDiffOneArgPushforwardPrep{SIG, <:JVPCache},
+        backend::AutoFiniteDiff,
+        x,
+        tx::NTuple,
+        contexts::Vararg{DI.Context, C}
+    ) where {SIG, C}
+    DI.check_prep(f, prep, backend, x, tx, contexts...)
+    # store the value f(x) inside the JVPCache since it will not change
+    copyto!(prep.f_in, f(x, map(DI.unwrap, contexts)...))
```
ChrisRackauckas: This isn't quite it. The problem is that we have already computed `f(x)`, so we want to reuse it in computations. I would expect the function to take in `fx = f(x)` under the contract that it is the correct `x`.

gdalle: That's exactly what this function does: it precomputes `f(x)` and stores it inside the `prep` while promising that `prep` will only ever be used at the same point `x` afterwards. What am I missing?

ChrisRackauckas: We have already computed `f(x)` outside of DI as the end point of an interpolation. We now want to give DI f(x) and then have it only compute `f(x+eps)` for the finitediff jvp

gdalle: I don't think that is easily done with the current DI API. What this PR does is provide you with a preparation function that computes `f(x)` for you, and tells `DI.pushforward` to reuse it. So it does waste one `f` call overall if you already have `f(x)` before calling `DI.prepare_pushforward_same_point`, but it then saves one `f` call per execution of `DI.pushforward` with the resulting `prep` object.
Can you point me to where in OrdinaryDiffEq we'd have to insert this?

ChrisRackauckas: For a Rosenbrock method for example, it would have already calculated this `f(x)` here: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl#L842. It would go into the reinit of `A` https://github.com/SciML/OrdinaryDiffEq.jl/blob/4bf7627b75e881a14613f1b708f106e4243a882c/lib/OrdinaryDiffEqDifferentiation/src/linsolve_utils.jl#L59-L63 since A is a matrix-free operator for the `J*v` operation, used in a Jacobian-Free Newton Krylov (GMRES) method to solve the Newton `Jx=b` without building `J`, so it's just represented as an operator https://github.com/SciML/OrdinaryDiffEq.jl/blob/4bf7627b75e881a14613f1b708f106e4243a882c/lib/OrdinaryDiffEqDifferentiation/src/operators.jl#L28-L73 that calls the prebuilt cache. What we previously had was a way to declare that `f(x)` was already calculated via a boolean state, and then `fx` would just need to be kept updated outside the differentiation system under the assumption the bool was set correctly. 

I noticed this was missing because the pre-DI code had this specialized with FiniteDiff.jl, that's why its cache build has this `fx` as something able to be passed, but then with the new solver method https://github.com/SciML/OrdinaryDiffEq.jl/pull/4017#issuecomment-5210726211 the benchmarks showed that our implementation was taking 1 extra evaluation per-step, leading to like a 20-25% performance reduction, and I tracked it down to being due to the change to DI which removed this handling.

gdalle: So if your `JVPCache` will only ever be used at one `u`, my thought would be to replace https://github.com/SciML/OrdinaryDiffEq.jl/blob/4bf7627b75e881a14613f1b708f106e4243a882c/lib/OrdinaryDiffEqDifferentiation/src/operators.jl#L138-L140 with `DI.prepare_pushforward_same_point`

ChrisRackauckas: We don't want this allocating, I assume we cannot assume `DI.prepare_pushforward` is non-allocating? (The fact that this code seems like it's allocating is a regression that I am concurrently handling 😅, it should be like a `reinit_pushforward!` or whatever DI function would update an existing cache to a new `u`).

gdalle: That would be `DI.prepare!_blablabla`, which I implement for SciML to be able to resize inputs dynamically and resize the cache to go along with it?

ChrisRackauckas: The Jacobian code uses it https://github.com/SciML/OrdinaryDiffEq.jl/blob/4bf7627b75e881a14613f1b708f106e4243a882c/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl#L376-L381. It passes in `uf`: does DI prevent the `f(u)=fu` call with that? I assume it probably has 1 extra f-eval as well from this, though that's probably harder to ever really matter since at 10x10 that's a 9% regression 🤷 ehh).

But I think the jvp-operator code did not get as much love during the update to DI. It's my mistake as we didn't have tests that were strictly guaranteeing the right number of `f` calls and non-allocations. It's not really the biggest deal because generally you only use Jacobian-Free Newton Krylov for very large systems, and so if the system is large then these allocations are "relatively" small, but after making a few other things 2x-3x faster this was now showing up as the last 20-25%. 

So yes, in a very concrete DI framing, what I am looking for is a way to give `uf = f(u)` in `prepare!_pushforward` so that I can change this operator code to reinitialize at new `u` in a non-allocating way, but also pass along the fact that I already have calculated `f(u)` in another part of the code.

gdalle: Ok so conceptually your workflow would look something like this?
```julia
prep0 = prepare_pushforward(f, backend, x0), (dx0,))
for i in loop_over_x  # ODE solve
    # your code computes yi = f(xi) somehow
    prep1 = prepare!_pushforward_same_point(f, prep0, backend, xi, (dxi,); y=yi)
    for j in loop_over_dx  # iterative linear solve based on the JVPs at xi
        pushforward(f, prep0, backend, xi, (dxi_j,))
    end
end
```
In which case I'd have to add:

- the function `prepare!_pushforward_same_point`
- the kwarg `y` in there

ChrisRackauckas: Precisely

gdalle: That will cost you one karaoke duet per feature

ChrisRackauckas: What's your duet style? Are you like a Chester to my Mike, or is this like a "me and my shadow" kind of situation? I've also got a pretty high voice, so we could also go off the deep end with summer nights or empire state of mind.

gdalle: Opened https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/1055 to keep track, still gonna merge this one because it does make sense

gdalle:
Benchmark tests pass locally, failure might be due to Codecov. Trying to run them again here, but I'll have to split out benchmarks from the Codecov-affected test workflow at some point

DifferentiationInterface/ext/DifferentiationInterfaceFiniteDiffExt/onearg.jl
ChrisRackauckas: We still need to keep track of getting a new karaoke DJ too so that you're allowed on stage. That definitely qualifies as an issue, just a non-technical issue.

gdalle: I think we're too late for this week tho

ChrisRackauckas: Yes, no worries. I travel around. You just need to be ready at any time. It can happen at any moment.

gdalle:
Superseded by #1061

